PR #18, opened 2020-11-10
Consolidated Ranges, added more pulldowns to tooltip

crnormand:


mjeffw:
I almost did the same thing (generate the text from the structs). Was planning to do so today, now I can focus on init.

PR #26, opened 2020-11-11
Feature/initiative

mjeffw:


mjeffw:
**Foundry BUG?** actor.sheet.rendered is 'false' if opened in response to double-clicking the token. In that case, the player will have to close and reopen his character sheet.

PR #33, opened 2020-11-14
Foundry specific GCS export/import and massive improvement in import speed

crnormand:
I've created a Foundry VTT export template for GCS and updated the import logic to accept it.   While still being backwards compatible to the older FG import.   

I found out that the DB doesn't delete keys from objects unless you use a special syntax.   If you don't, it will just merge objects... which means that on a second import, you will still have all of the original data (unless it is overwritten by using the same keys).   Which means, that is you remove an item, like a piece of equipment...  and re-commit the list with one item missing... it won't go away.

So I had to preface all of my list updates like this:
await this.update({"data.-=ads": null});
await this.update({data.ads": list}};

And the import was literally taking seconds... you could see the character sheet shuddering as various lists updated.   It was bad.  @Stas mentioned batching up all of the updates, so you would be proud of me, I used some new syntax.

I collect up all of the changed data:
			commit = {...commit, ...this.importAttributesFromCGSv1(c.attributes)};
			commit = {...commit, ...this.importSkillsFromGCSv1(c.abilities?.skilllist, isFoundryV1)};
			commit = {...commit, ...this.importTraitsfromGCSv1(c.traits)};
Using the spread operator "...", and then I filter to get the deletes and then the adds:

		let deletes = Object.fromEntries(Object.entries(commit).filter(([key, value]) => key.includes(".-=")));
		let adds = Object.fromEntries(Object.entries(commit).filter(([key, value]) => !key.includes(".-=")));
		
		await this.update(deletes);
		await this.update(adds);
 And now imports are instantaneous again ;-)


mjeffw:
I use the spread operator in Dart; they probably borrowed the syntax from javascript. or vice versa.

PR #51, opened 2020-11-26
Massive update to GCA import

crnormand:
Also, will either show "appearance" from GCA or "eyes", "hands", etc. from GCS


crnormand:
New export utilities in "exportutils" subdirectory

PR #67, opened 2020-11-30
Bunch-o-stuff

crnormand:
Started support for blindroll sematics.   So GM can ask everyone for a Perception roll and they can't see it.
Updated parselinks to handle 4d+1X4 (0…5) damage.  Optional space allowed in front of armor divisor (but not multiplier).
If formula is NOT a damage roll, pass to Foundry without armor divisor, and convert "X" and "x" to "*".  4d+1*4
Warn user if they try to import using a Fantasy Grounds export (and yes, someone installed 0.8.0, didn't read the docs and DMed me on Discord as to why it didn't work).     So now it shows an error, and then puts the URL in the chat log.


mjeffw:
FYI, I removed support for armor divisor with a space in front of it in the damage chat code. I can add it back.

PR #71, opened 2020-11-30
Fix null actor on damage roll when no LastActor

crnormand:


mjeffw:
We may have to support a null actor for things like traps or adverse climate damage, eventually.

mjeffw:
I propose we send an actor or a string that describes the source of the damage.

PR #76, opened 2020-12-02
defining new css file and using it as class

gtludwig:


gtludwig:
I'll have to rename this later, but might give you a good idea what I would like to do

gtludwig:
BTW, before approving / merging, can we have a quick chat about this?

crnormand:
The question is, does this work?   I didn't think Foundry automatically reads CSS files from the "styles" directory.    I have seen most other projects use some kind of 3rd party tool (like less) to split up their CSS into manageable chunks, and then combine it into a single CSS for Foundry.  But I am NOT a CSS (or a Foundry expert).

mjeffw:
I'm not a web developer either, but I generally don't like to hardcode sizes into things, generally preferring to use relative measures (like % or fr). Also;  I like to name the container and then use child selectors to individually size the sub-elements.

crnormand:
I found out that we can separate our CSS, so long as we identify it in the system.json, so we CAN break it apart (and we have started to do so).
As for that particular html, it is a direct copy of the HTML export that GCS creates..   If Rich changes his HTML output and we want to use it, we will either need to copy it whole again and reapply all of our framework changes or we will have to go through it manually to figure out how the HTML has changed and then apply the CSS/HTML changes.    Not a good situation in either case.
Finally, I agree wholeheartedly with Jeff... I am learning this whole CSS stuff, and I can certainly see the advantages.

gtludwig:
Good stuff, lads!
I'm happy to continue with my implementation of this.
I was unable to put in any time into it the last few days because of boring
stuff like people actually paying me for my time... Hehe!

gtludwig:
Took me a little while to start contributing again, but here I am!!!

PR #140, opened 2020-12-24
Implemented range ruler support for grid units other than yards and refactored range code

Exxar:
Ok some explanation for what I did :D

The code should be slowly reorganized. A `lib` folder is usually only used for various external libraries. Project code usually goes elsewhere. In order to not shake up the existing organisation too radically, and also according to what I've seen is often done in other systems and modules, I suggest moving the project code to the `module` folder. There it should be organized in folders according to the "problem domain", so I've moved this code to the size-speed-range folder. Each problem domain should best be split into a pure "gurps logic" namespace, and another one where everything is tied together with Foundry, so that those different concerns are not mixed. But code that does the same thing should also be located together, which is why this split is better on the individual problem domain level than on the global level. So I've gone ahead and split it into `domain` and `foundry` folders. `foundry` could theoretically be further split into Foundry infrstructure (hooking stuff up via hooks, chat commands, sheets etc) and use case / UI concepts which can be represented independently of the underlying Foundry infrastructure itself, but I'd leave this for later, if at all.

The most important piece of refactoring in ranges.js was I think refactoring the monsterHunter2Ranges and basicSetRanges objects. They were problematic because they were both directly used for internal calculations of range penalties for the range ruler, and exposed publicly to various consumers for their own purposes. There was also the need for a note saying that they have to be kept in order for the ruler calculation to work. This made the code extremely fragile and changing it could have far-reaching unintended consequences. Which is why I rebuilt it like I did - now all range calculations are handled by the classes encapsulating the two tables, while the tables as structured data for the use by external consumers are "sourced" from those classes but otherwise not used for any internal logic. The responsibilities are better separated within the range module, and external consumers are also better separated from the internal logic.

I'll also comment some specific changes I made to the code in a couple places. Also some "global" remarks:
- Consistent indentation throughout the project is important :) 4 spaces is generally the most common one. It was all over the place in ranges.js
- I've also added semicolons to end of all statements, that was likewise all over the place in the file. You need to decide on a standard for this and then stick to it; I suggest using semicolons :P
- Speaking of a code style standard, I suggest using [Prettier](https://prettier.io/). It is a plugin for editors/IDEs (supports both VS Code and Eclipse) which takes care of code formatting for you according to its (very reasonable) opinions. That would allow you to just not worry about code style (though in some IDEs it must be manually run).
- if/else and similar clauses should always have braces around their code even if it is just one statement, for better readability


module/size-speed-range/foundry/ranges.js
```diff
@@ -0,0 +1,180 @@
+'use strict';
+
+import {displayMod} from '../../../lib/utilities.js';
+import {SSRT} from '../../../lib/gurps-foundry-ssrt-lib/src/js/SSRT.js';
+import {RangeBandTable} from '../domain/RangeBandTable.js';
+import convertToYards from '../../../lib/gurps-foundry-ssrt-lib/src/js/UnitsOfMeasure/Conversion.js';
+import parseDistance from '../../../lib/gurps-foundry-ssrt-lib/src/js/Parsing.js';
+import * as conversionMethods from '../../../lib/gurps-foundry-ssrt-lib/src/js/UnitsOfMeasure/Methods.js';
+
+const SYSTEM_NAME = 'gurps';
```
Exxar, line 10: This and a couple others were previously exported; don't export variables which will not be used in other files.

module/size-speed-range/foundry/ranges.js
```diff
@@ -0,0 +1,180 @@
...
+
+            /* Define Settings */
+            game.settings.register(SYSTEM_NAME, SETTING_NAME, {
+                name: 'Default range modifier strategy:',
+                hint: 'Sets the formula to use to calculate range penalties.',
+                scope: 'world',
+                config: true,
+                type: String,
+                choices,
+                default: STRATEGY_STANDARD,
```
Exxar, line 99: Always use constants instead of strings/numbers if they represent a distinct concept which is referred to in multiple places.

module/size-speed-range/foundry/ranges.js
```diff
@@ -0,0 +1,180 @@
...
+  - Maintains an instance variable (rangeTable) that contains the current set of
+    range bands based on the chosen strategy. This table is to be used only for display
+    and interactive roll mod application purposes
+
+  - Maintains an instance variable (bucketModifiers) that contains an array of
+    modifier text for the modifier bucket.
+ */
+export default class GURPSRange {
+    constructor() {
+        this._setup();
```
Exxar, line 77: I suggest adopting the convention to prefix the names of all private methods and class properties with _. This is what Foundry does, and is a useful thing to do in JS where there is no real private/public enforcement.

module/size-speed-range/foundry/ranges.js
```diff
@@ -0,0 +1,180 @@
...
+                    conversionMethod: conversionMethods.real
+                })
+            );
+        }
+
+        return SSRT.speedRangeFromExpression({expression: `${distance} ${units}`});
+    }
+
+    async _updateAfterSettingsChange() {
+        const currentStrategy = getStrategy();
```
Exxar, line 167: variables should always be named to accurately express their purpose; currentStrategy is much better than currentValue in that regard

Exxar:
@mjeffw @crnormand Pinging you guys for a review as well as some comments I made, no idea what the notification settings here are

module/size-speed-range/foundry/ranges.js
mjeffw: Yep, I was doing that on the applydamage.js and damagecalculator.js.

mjeffw:
Chris and I need to talk about these changes before merging. There's a lot that I agree with, but ultimately this is Chris' repo. Plus this is a bigger set of changes without prior input that I'm comfortable with. I was literally about to work on a large refactor (coincidentally implementing some of the same changes) and this makes me pause.

I thought you were going to work on a drop in replacement for the ADD for conditional injury?

crnormand:
What Jeff said.  We can't accept huge wholesale changes to the codebase.  

I did the same thing when I started contributing to GCS.  I was so used to owning the code, and being the most experienced dev on the team (I've been using Java for 25 years), that I made large changes throughout Rich's code and made one large PR.     

He rightly rejected it 😜

My suggestion to you is to make lot of small pull requests, focusing on a single feature/function for each.  

Thanks for the suggestion on the prettifier.   Eclipse is giving me fits with its odd indenting.  

Overall, I agree with your refactoring suggestions.   Jeff and I have lots of experience in this area.   You have to remember, this code was hacked together in 8 weeks, all while learning the language.  I took tons of shortcuts trying to get functionality out the door.   I didn't spend a lot of time refactoring code. 

When I get back to my computer (maybe sometime after Christmas), I can go over all of the changes and we can discuss them further.

Exxar:
> I thought you were going to work on a drop in replacement for the ADD for conditional injury?

Yeah, but then I opened the code and this caught my eye and I felt a need to do something here. Sorry for talking about it with Chris over DMs and leaving you out!

> P.S. Chris and I are both senior developers with a lot of experience, but we have slightly different philosophies on writing code. I tend towards lingering over code to make it readable/extendible and "pretty" in the editor. Chris likes to get functionality out the door and will refactor later. As evident by our output -- I did the ADD and a few smaller contributions, everything else is Chris (mook generator, alternate character sheets, OtF rolls, range ruler, etc. etc.).

Excellent, we can gang up on Chris with our similar preferences then 😜

Anyway, I primarily made this PR as a case study to show my preferences and suggest where to go. I don't expect it to be merged without a good prior discussion and deliberation. I haven't implemented anything significant anyway.

I'd like to contribute here, because this seems to be the system I'll be using for my own campaigns and contributing seems like a better idea than keeping my own private fork. We'll see how it goes. I don't want to promise any commitments but I also don't want to push my own preferences and wishes, because in the end this is Chris' project and I'm already the third (if not 4th etc) chef in the kitchen :D

> You have to remember, this code was hacked together in 8 weeks, all while learning the language. I took tons of shortcuts trying to get functionality out the door. I didn't spend a lot of time refactoring code.

Yes, absolutely, this is the reason why you have a working system and I don't 😁. So just go go go! But unfortunately, after a day of expending  my energy and enthusiasm at work, I can't muster the motivation to work on code that isn't nice and pretty. It doesn't bring me joy. So if I'll contribute I won't be able to resist cleaning up code I touch. Which is why I'll want to sync up properly with you guys before doing it.

PR #150, opened 2020-12-31
Proof of concept for Conditional Injury

Exxar:
This is just an initial proof of concept to run it by you guys and see what you think, it is not complete and should not be merged.

I have implemented the basic bookkeeping functionality necessary to run games with Conditional Injury (Pyramid #3/120). There is a system setting which toggles CI on and off, and for this proof of concept I've only made modifications to the full sheet and the editor sheet. On both of those sheets the HP tracking fields are replaced by the injury fields.

Full sheet:
full_sheet

Editor sheet:
editor_sheet

I've implemented the following functionality on the full sheet:
- Injury severity and days to heal can be manipulated in the same way as any other resource tracker
- Manipulating severity automatically sets days to heal to the appropriate value for the new severity (each injury severity has a certain number of days which are required to pass before the injury is fully healed)
- Manipulating days to heal automatically sets severity to the appropriate value for the new days to heal (this field is intended to be used when natural or magical healing occurs; the user reduces the days to heal by however many days of healing were applied, and if the days to heal threshold for a lower injury severity is reached, the severity field is updated automatically)
- The reset button puts an empty string into the severity field and sets day to heal to 0 (severiities are counted from -6 to 6, so 0 actually means the equivalent of being at 0 HP which is why the severity field can't use 0 for "no injury")

The intended workflow here is adjusting the severity field as the character is injured, and adjusting the days to heal field as the character is healed.

The editor sheet just has editing of the relevant fields, without any calculations being performed by the sheet. The user can input whatever they like in the fields.

Unfortunately, Foundry input field data binding is causing me problems when days to heal are directly edited on the full sheet. When days to heal input field is directly edited, instead of the severity field being adjusted to the appropriate value once the days to heal field loses focus, the new (appropriate) value for the severity is only briefly flashed for a split second in that field before reverting to the previous value (the days to heal value is appropriately persisted it seems). So far I haven't been able to fix this and I'm outta ideas. Curiously, doing it in the other direction, so directly editing the severity input field, correctly sets the value of the days to heal field without such issues.

So unless you have an idea why this is happening, I was considering making the days to heal field read-only, and only able to be manipulated using the spinner buttons. This could however be somewhat impractical because often times many days of healing are applied at once and having to click 10 times isn't good UX. Another idea would be to provide a button which opens a popup where the  user can enter the number of days of healing to apply but that's not the way other trackers are manipulated.

So anyway, take a look and tell me what you think @crnormand @mjeffw

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .DS_Store
 .project
 Thumbs.db
+scripts/foundry.js
--- a/README.md
+++ b/README.md
@@ -1,11 +1,10 @@
 ### Users Guide for [GURPS 4e game aid for Foundry VTT](https://bit.ly/2JaSlQd)
 
-# Current Release Version 0.8.3
+# Current Release Version 0.8.4
 [If you like our work...](https://ko-fi.com/crnormand)
 
-<a href="https://ko-fi.com/crnormand">
-<img height="36" style="border:0px;height:36px;" src="https://cdn.ko-fi.com/cdn/kofi2.png?v=2" border="0" alt="Buy Me a Coffee at ko-fi.com">
-</a>
+<a href="https://ko-fi.com/crnormand"><img height="36" src="https://cdn.ko-fi.com/cdn/kofi2.png?v=2"></a>
+
 
 To install the latest release, use this manifest URL:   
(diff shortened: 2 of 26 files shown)


Legendsmith:
Awesome thank you!

crnormand:
Ok, I have merged this onto my branch and will work on it.    I need to re-read the article to completely understand the issues...but we will get there.   Thank you for the excellent work!

PR #171, opened 2021-01-12
hit location for i18n

gtludwig:


crnormand:
Ok, got a chance to look at it.    I'm going to reject it mainly because all of the hitlocation stuff is now in module/hitlocation/hitlocation.js and actor.js.      And while I normally agree with removing "magic" strings, we wouldn't be saving much effort here.    Since the actual names of hit locations are really only important during the import, and after that, they are just displayed.   So there is no code in the rest of the system that refers to "LEFT_ARM".    A static for this would only be using during the text matching of the import... and during that, I would prefer see the actual text, instead of referencing back to a list of static variables.

gtludwig:
OK, not a bother! I have only myself to blame - I took very long to resolve
the merge conflicts. Thanks for the feedback!

Gustavo Tomazi Ludwig
gustavo.ludwig@gmail.com


<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Virus-free.
www.avast.com
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

PR #198, opened 2021-01-22
Feature/css upgrade

crnormand:


crnormand:
For the most part, the did not provide a lot of re-use, so there didn't seem to be a reason to remove the style attributes from the HTML elements.   Yes, there were a few cases where some fields were able to use the same CSS class, but outside of that one HTML, I didn't see the benefit of having such  specific css classes, such as "w9px".

Now, your work on this matter was in the actor-sheet.html... which we don't actually use.   It was the original test html.    And honestly, I wouldn't recommend trying the same thing on the actor-sheet-gcs.html.    Mainly because this is a direct copy of the the GCS HTML output, and if Rich makes a change to his HTML, it would be much harder to detect the differences, if we have replaced all of the style attributes with css classes (because basically every line would show as a difference).

One final note on this... it seems that in css_table.css, you define rules for "tr" and "td", and as I understand it (and trust me, I am not an expert on this), that seems to change the rule for ALL "tr" and "td" elements in our game system, no matter which html file they are in.    It may not be showing up, because all other instance of  "tr" are being override by more specific CSS... but I don't think we want to define defaults for such high level html elements.   But then again, I am not an HTML/CSS expert by any means.

PR #223, opened 2021-01-29
i18n continued

gtludwig:


crnormand:
I noticed that the GCS copyright isn't correct:

  <div id="footer">{{localize "GURPS-copyrightGCS"}}
  
  Shouldn't it be:
  
    <div id="footer">{{localize "GURPS.copyrightGCS"}}

PR #228, opened 2021-01-30
i18n updates

gtludwig:


gtludwig:
This is a bit trickier to test because it depends on damage rolls. I am following the same pattern as the previous PRs, so I think it is fine. Let me know if I missed anything.

Issue #229, opened 2021-01-30
GCS Item Values over 1000 are being set to 0


trev-dev:
Just noticed this after updating to 0.8.9 - any item in the inventory that's worth more than $1000 is being imported at $0. I checked the XML and the numbers are as they should be there.

If I go into the XLM and drop the comma on the thousands separator, the import works fine.

Either GCS needs to avoid inserting commas our the import parser needs to prune them.

crnormand:
Can't beat that... the person who makes the issue, also provides the bug fix!   Thanks!

trev-dev:
At some point a code refactor happened that reversed the fix for this issue.

crnormand:
My bad.   I didn't move the comma check when refactoring.    It as been fixed https://github.com/crnormand/gurps/pull/305

PR #231, opened 2021-01-30
sanitize input for cost values over 999

trev-dev:
A cheesy/quick fix for sanitizing input values for the `cost` property of Equipment. This is mainly because the check performed here evaluates strings with commas as `NaN`:
https://github.com/crnormand/gurps/blob/6bb715fc98f8183b113880da16888660e8edaf33/module/actor.js#L1223

The end result of this bug was that all equipment imported that was valued at over $999 was being treated as 0.

Related issue: https://github.com/crnormand/gurps/issues/229

### DIFF
--- a/module/actor.js
+++ b/module/actor.js
@@ -501,7 +501,7 @@ export class GurpsActor extends Actor {
         let eqt = new Equipment();
         eqt.name = t(j.name);
         eqt.count = i(j.count);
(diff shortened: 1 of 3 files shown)


trev-dev:
Closed as I meant to branch my commits/pull requests. Oops!

PR #233, opened 2021-01-30
Fix logo aspect ratio and adapt navigation

trev-dev:
I noticed that the GURPS logo was compressed when it was added in the last update so I made some tweaks to fix that. Nothing personal but JQuery makes me itch, so I swapped that out in gurps.js for perfectly good JS.

Here's what the update looks like:
Screenshot_20210129_220313


crnormand:
As mentioned in my DM... I included the fix (in spirit), so I won't be merging this PR.

mjeffw:
This looks good! Even through this was a very small change, could you create a branch and submit a PR next time?

PR #249, opened 2021-02-02
i18n updates

gtludwig:


crnormand:
Wow, that's a lot of work... but I trust you (plus, I know where you live... ok... I know which island you are on, and really, how big can it be?... but getting back to the trust issue, I am going to accept this PR)

PR #263, opened 2021-02-05
i18n for ADD

gtludwig:


gtludwig:
@crnormand, @mjeffw and @trev-dev,

This is the initial PR for the ADD i18n.

It is a bit big because of replacing the values in English with placeholders and adding the terms to both `lang/en.json` and `lang/pt-br.json`. 

I am happy enough with this, but it is not release-ready yet.

The hit-location and damage type list and a few other place holders come from javascript files I did not want to touch in this PR.

Please take your time reviewing this.

Cheers,
Gus

PR #272, opened 2021-02-06
Feature/explosion damage

mjeffw:
Added support for Explosion damage (B414) to the ADD.


lang/pt_br.json
```diff
@@ -121,7 +121,7 @@
   "GURPS.addHardenedDRLevel" : "RD Ampliada Nível",
   "GURPS.addInjuryTolerance" : "Tolerância a Danos",
   "GURPS.addInjuryToleranceUnliving" : "Não vivo",
-  "GURPS.addInjuryToleranceHomogeneous" : "Homogêneo",
+  "GURPS.addInjuryToleranceHomogenous" : "Homogêneo",
```
gtludwig: Thanks for correcting the typo!

mjeffw: No problem. Both are English words. 'Homogenous' and 'homogeneous' pretty much mean the same thing, but GURPS uses the former, not the latter, in the rules.

PR #293, opened 2021-02-10
Rewrite and unify damage parsing

crnormand:
Support non-dice damage [3 cut].   Can support all other modifiers.   This is a legal (if stupid) roll formula:  '5-2x3(2) cut', and it displays:  
5-2 = 3.
Total = 3 x 3 = 9
If there are no dice rolled, and there are no other modifying factors to the damage "roll", the drag-able section does not contain a collapsible section (since there is nothing to display).
Don't tell anyone... we now support any sided dice for non-targeted (e.g attribute, skill), non-derived damage (e.g. SW+1) rolls.    Examples: [3d4], [2d10 cut], [2d100-30x3(0.5) imp]
We only show the dice sides if they are part of the damage formula (e.g. '3d6').   If they are not included, we assume 6 sided dice, and we display the roll in 'GURPS' format (e.g '3d')

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ This is what we are currently working on:
     - Added Mook Generator defaults editor
     - GCA export/import now handles parents for Ads/Disads/Spells and Skills
     - Rewrite of damage parser, more uniform handling of multipliers/divisors, etc.  support for const damage [1 cut]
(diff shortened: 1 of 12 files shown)


crnormand:
Also changed Knockback chat message to use [DX | S:Acrobatics | S:Judo] OtF form so the player doesn't have to figure out which is best

PR #303, opened 2021-02-13
sanitize string numbers over 999

trev-dev:
https://github.com/crnormand/gurps/issues/229 - a regression has caused this bug to crop up again. I have sanitized the cost + weight imports for equipment items by dropping the commas.


crnormand:
Actually, that was my fault.    I moved the cost/quantity/weight handling into a static Equipment method since we can get equipment information from other sources besides the import (user edit and Mook generator), and I forgot to include the comma fix.

PR #343, opened 2021-02-26
i18n pt-br text updates

gtludwig:


lang/en.json
```diff
@@ -192,7 +192,7 @@
   "GURPS.hitLocationWhere": "Where",
   "GURPS.hitLocationPenalty": "Penalty",
   "GURPS.hitLocationDR": "DR",
-  "GURPS.hitLocationDRWithDivisor": "Effective DR",
```
mjeffw, line 195: Why remove this line? I changed the text to "Effective DR" because after I added shotgun damage, this line showed the effects of applying armor divisor and/or the shotgun damage multiplier. So it should stay in the file.

mjeffw (changes requested):
Just need clarification on the one comment I left in this review.

lang/en.json
gtludwig: I removed that line by mistake. Re-added!

mjeffw (approved):

PR #367, opened 2021-03-03
javascript refactoring for bucket

gtludwig:
Slicing down the bucket modifiers into bite-sizes for improved maintenance

### DIFF
--- a/module/bucket/modifiers/sizemodifier.js
+++ b/module/bucket/modifiers/sizemodifier.js
@@ -1,0 +1,28 @@
+'use strict'
+
+export const sizeModifier = [
+    'Size Modifier (melee diff, ranged abs)',
+    '-10  Size 0.05 yard (1.8")',
+    '-9  Size 0.07 yard (2.5")',
+    '-8  Size 0.1 yard (3.5")',
+    '-7  Size 0.15 yard (5")',
+    '-6  Size 0.2 yard (7")',
+    '-5  Size 0.3 yard (10")',
+    '-4  Size 0.5 yard (18")',
+    "-3  Size 0.7 yard (2')",
+    "-2  Size 1 yard (3')",
+    "-1  Size 1.5 yards (4.5')",
+    "+0  Size 2 yards (6')",
+
+    "+1  Size 3 yards (9')",
+    "+2  Size 5 yards (15')",
+    "+3  Size 7 yards (21')",
+    "+4  Size 10 yards (30')",
+    "+5  Size 15 yards (45')",
+    "+6  Size 20 yards (60')",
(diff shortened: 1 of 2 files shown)


gtludwig:
This is part of the i18n drive.
I'm sure it's some stupidly simple, but I might need to chat to some of ye to get it working!

crnormand:
What are you wanting to accomplish here?

Are you just trying to see if you can move the static variable into another .js file and import it?

My personal thinking, and this is just my opinion and may not be shared by the others, but I don't see any value in moving the static variable declaration into a separate .js file.   For me, it just hides the code.   At some point in the future, this will hopefully become an editable field that the GM can use to customize the modifiers list.   All of these will be.   So moving them into separate .js files doesn't buy us anything.

gtludwig:
I am hoping to accomplish two things here:
1. Ease of maintenance - I believe smaller files are easier to develop and maintain.
2. Decoupling of static fields into another .js file and import the internationalized version.
This is intended mainly as an exercise for achieving i18n support.

crnormand:
Your 1st point is a tricky subject.   I could argue that really small files do not make it easier to develop and maintain.   Instead of a developer having to find something in 30-40 files, they would have to search through hundreds.   

My metric is to see if the code in question stands by itself.   If you asked it what it did... would it be a useful thing?  Or is it just in support of something else.   In this particular case, the static string does nothing by itself.   Its sole purpose in life is to populate the Modifier Bucket tooltip.

I understand that you have segregated the files into another folder.   But that gets back to my first point.   I do not see the value of having a lot of small .js files containing the static initializers, such that someone working on that code would have to have multiple editors open to be able to "see it all".

In an example, consider a developer who wants to change a string (or internationalize it).   We might use a similar phrase in various modifiers, and we want to change/fix it in all of the modifiers.   If all the code is in different editors, you have to move between them all to find the right spot, whereas, if they were in a single file, you would only have to scroll to the correct spot.   A very minor reduction in the number of actions required to edit... but it is there, none the less.   And multiplied by hours and hours of development...  well, that's my point.

I fall back on the "ask the code what it is doing" metric... and if it is a "big enough" thing to stand alone, then do it.   Of course, this is HUGELY subjective... and I can only go by my 30+ years of experience as a developer.

Concerning point number 2, what is the value in decoupling the static initializers individually?   They serve one purpose... to statically initialize the tooltip.   I only look to decouple code when there are 2 'subsystems' that each have their own purpose, yet are tied together.    

So, my suggestion would be to move ALL of the static initializers into a single separate .js file (not one for each variable).   Any internationalization work can be done there.   The main modifier.js does not need to have the initializers there (although, I don't think it hurts anything).

But maybe the others have different opinions.

PR #384, opened 2021-03-07
Implemented selectable armor divisor in the apply damage dialog

Exxar:
@mjeffw we talked about this a loooong time ago. Appears to be working, except the 100 and infinite levels of armor divisor break the layout.

### DIFF
--- a/templates/apply-damage/apply-damage-dialog.html
+++ b/templates/apply-damage/apply-damage-dialog.html
@@ -171,11 +171,28 @@
             <div><label for='tactical-blunttrauma'>{{localize "GURPS.bluntTrauma"}}</label></div>
             <div class='pdflink'>B379</div>
 
-            <div><input id='tactical-armordivisor' type='checkbox' value='armordivisor' {{#if
-                CALC.useArmorDivisor}}checked{{/if}}></div>
-            <div><label for='tactical-armordivisor'>{{localize "GURPS.addArmorDivisor"}}</label></div>
-            <div class='pdflink'>B378</div>
+          </div>
 
+          <div class='shaded collapsible-wrapper options-drawer'>
+            <input id='tactical-armordivisor' class='toggle' type='checkbox' value='armordivisor' {{#if
+                   CALC.useArmorDivisor}}checked{{/if}}>
+            <label for='tactical-armordivisor' class='label-toggle'>{{localize "GURPS.addArmorDivisor"}}: <span
+                    class='pdflink'>B378</span></label>
+            <div class='grid grid-5col continuation collapsible-content'>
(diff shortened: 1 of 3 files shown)


templates/apply-damage/apply-damage-dialog.html
```diff
@@ -171,11 +171,28 @@ <h4>{{localize "GURPS.addTacticalRules"}}</h4>
...
+            <label for='tactical-armordivisor' class='label-toggle'>{{localize "GURPS.addArmorDivisor"}}: <span
+                    class='pdflink'>B378</span></label>
+            <div class='grid grid-5col continuation collapsible-content'>
+              <label><input type='radio' name='tactical-armordivisor' value='2' {{#if (eq CALC.armorDivisor 2)}}
+                            checked{{/if}}>&nbsp;2</label>
+              <label><input type='radio' name='tactical-armordivisor' value='3' {{#if (eq CALC.armorDivisor 3)}}
+                            checked{{/if}}>&nbsp;3</label>
+              <label><input type='radio' name='tactical-armordivisor' value='5' {{#if (eq CALC.armorDivisor 5)}}
+                            checked{{/if}}>&nbsp;5</label>
+              <label><input type='radio' name='tactical-armordivisor' value='100' {{#if (eq CALC.armorDivisor 100)}}
```
mjeffw: Shouldn't this value be 10? (from B102).
```suggestion
              <label><input type='radio' name='tactical-armordivisor' value='10' {{#if (eq CALC.armorDivisor 10)}}
```

templates/apply-damage/apply-damage-dialog.html
```diff
@@ -171,11 +171,28 @@ <h4>{{localize "GURPS.addTacticalRules"}}</h4>
...
+                            checked{{/if}}>&nbsp;100</label>
+              <label><input type='radio' name='tactical-armordivisor' value='-1' {{#if (eq CALC.armorDivisor -1)}}
```
mjeffw, line 190: Your PR mentions 'infinite'. Is that what value -1 is equivalent to? Where is an 'infinite' armor divisor defined?

mjeffw (changes requested):
See my two comments on AD 100 and -1.

Exxar:
Umm, `damagecalculator.js` line 24 defines 
```
// -1 means 'Ignores DR'
const armorDivisorSteps = [-1, 100, 5, 3, 2, 1]
```
This is the division progression the existing DR hardening handling code uses. So that's where -1 as infinite comes from. I also noticed that divisor 10 is missing, but I didn't want to do anything except to introduce this new feature initially. If you're otherwise fine with this PR I'll gladly extend this to include armor divisor 10.

PR #394, opened 2021-03-07
Implemented support for Injury Tolerance (Damage Reduction) in ADD

Exxar:
This is a form of Injury Tolerance introduced in GURPS Powers and used extensively in GURPS Supers. It divides injury by a specific number. I've implemented it to be separate from the existing Injury Tolerance values in the ADD because you can both have Injury Tolerance (Unliving/Homogenous) and Injury Tolerance (Damage Reduction) and they are cumulative. Since Injury Tolerance (Damage Reduction) comes in infinite levels (divides injury by 2, 3, 4, 5, 6 etc) I've implemented it as a numeric input field instead of a radio selection or similar.

### DIFF
--- a/module/damage/applydamage.js
+++ b/module/damage/applydamage.js
@@ -238,6 +238,29 @@ export default class ApplyDamageDialog extends Application {
       .find('input[name="injury-tolerance"]')
       .click(ev => this._updateModelFromRadioValue($(ev.currentTarget), 'injuryToleranceType'))
 
+    // if checked, target has Injury Tolerance (Damage Reduction)
+    html
+        .find('#damage-reduction')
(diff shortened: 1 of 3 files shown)


mjeffw:
I already got yelled at by @crnormand for merging your previous PR directly into `main`. ;-) Could you make a PR to merge into `feature/damage-reduction` instead?

Exxar:
You got it!

mjeffw:
Thanks, I will take a look and merge it into main if there are no concerns. It may be a while before it makes a release, we'd doing a big experimental refactor.

Exxar:
I can use main or even this branch anyway hehe.

PR #695, opened 2021-05-25
Converts all png and jpg to webp, changes all links to lead to webp, …

xdy:
…removes jpg and png files.

While doing this I noted that npc-input.css points to two missing files (ui/parchment.jpg and ui/parchment-white.jpg), I still changed the link to point to an equally nonexistent webp file.

Approximately halves the size of the system (not all that significant given that the zip is <3mb, but should help at least a little for those playing with slow/flaky internet.)


crnormand:
Oh, cool, thank you!   We will merge it AFTER we get this dang 0.8.X migration done.

xdy:
Probably a good idea. :)

PR #697, opened 2021-05-26
changed templates from .html to .hbs

danielrab:
handlebars templates are supposed to have .hbs extension. There were two .hbs templates in the project but the rest were .hbs <br>
using .hbs over .html enables syntax highlighting, reducing the probability of mistakes <br>
to verify that I didn't forget any of them, I searched for all the instances of "html" and ensured that none of them are related to the templates, however additional verification would be nice.


danielrab:
things have changed since then, so this needs to be redone

PR #771, opened 2021-06-12
Equipment Library Import

rinickolous:
Added the ability to import GCS equipment libraries directly into Foundry.
What this currently does:
- Imports the basic equipment info.
- Imports melee and ranged weapons.
- Imports the following GCS item features:
  - Attribute bonuses (except for Basic Speed and Basic Move, which seem to be unsupported for OTF).
  - DR bonuses.
  - Skill and spell bonuses, as well as attack bonuses.
  - Weapon damage bonuses*.

Other features (reaction modifiers, conditional modifiers, skill/spell point bonuses, *weapon bonuses by required skill, attribute cost reduction, and contained weight reduction) don't seem to be currently possible to implement using OTF formulas.

All imported equipment is currently imported and displayed as a flat list (afaik, Foundry doesn't support items containing other items by default anyway). I tested the import function with DFRPG Equipment as well as the Basic Set. With those two, no obvious errors were produced, but some things may have slipped through the cracks.
The import function seems to screw up Unicode characters (e.g. "ð" is displayed as "Ã°"). I'm not sure how to fix this, but would like to fix it before committing.

### DIFF
--- a/lang/de.json
+++ b/lang/de.json
@@ -572,5 +572,7 @@
   "GURPS.toAllDXBasedRolls": "auf alle DX-basierten Würfe",
   "GURPS.update": "Aktualisieren",
   "GURPS.usage": "Nutzung",
-  "GURPS.yards": " yards"
+  "GURPS.yards": " yards",
+
+  "GURPS.itemImport": "Artikelsammlung Importieren"
 }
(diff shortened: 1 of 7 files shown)


rinickolous:
Forgot to mention:
GCS has a "+X damage per die" field for weapon damage, which is formatted in a way NaN GGA doesn't like (e.g "thr+1 (+3 per die) cr". As such, any damage fields of this type don't work properly.

Issue #765, opened 2021-06-11
Item image on hover cut off


LordOHelmet:
![item hover](https://user-images.githubusercontent.com/80178910/121737142-ca51f100-caf8-11eb-8016-43f0259a8e9b.JPG)

The item images shown on hover are great!
But the layout isn't perfect - yet.

crnormand:
LOL, tell me about it.   Hopefully someone will pretty up the CSS.   I made certain the src path for the icon is available, basically the "plumbing".   Making it look good was never my strong suit.

LordOHelmet:
I love that it is implemented after all.

By the way. There is no item preview on hover in the "Other Equipment" section and on the "Inventory Only" sheet.

3darkman:
I'm taking a look at it

crnormand:
It turns out this didn't quite cut it, @3darkman   It works fine if the item (or feature) is at the top of the character sheet, but if you have to scoll down to see it,  the image appears off the bottom of the character sheet.   My mouse is at the red arrow

![image](https://user-images.githubusercontent.com/8931036/121836134-e6909200-cca0-11eb-8ada-8db002855a0e.png)

3darkman:
urgh!!! I'll take a look later today to sort this out, I forgot to confirm it on the tab without tabs.

PR #775, opened 2021-06-13
small adjustment for the sheet with tabs to avoid the appearance of two scroll bars

3darkman:
I added a little css hack to avoid the automatic calculation of the tab size in the SetPosition function of the actor-sheet.js file


3darkman:
I also made some adjustments to the css to resolve the issue https://github.com/crnormand/gurps/issues/765

PR #787, opened 2021-06-16
rename sheet-tabs to gurps-sheet-tabs

danielrab:
sheet-tabs is a class already used by the core software, using it leads to all sorts of unexpected behavior including the issue with the tabs of the game settings not being centered properly.


crnormand:
Thank you so much for digging into this.  When Plape showed it to me I had NO IDEA what I would need to do.... so you don't know how much I appreciate this!

PR #794, opened 2021-06-19
Basic Quintessence Support and NPC CI Sheet

Legendsmith:
Adds quintessence support, and a Conditional Injury version of the NPC sheet but I'm not sure if that works.
Also adds the (4) armor divisor, for Survivable Guns


mjeffw:
Who's doing your language translations? My volunteers have been missing somewhat these days.

module/gurps.js
```diff
@@ -1516,6 +1518,7 @@ Hooks.once('init', async function () {
   GURPS.initiative = new Initiative()
   GURPS.hitpoints = new HitFatPoints()
   GURPS.ConditionalInjury = new GURPSConditionalInjury()
+  GURPS.Quintessence = new GURPSQuintessence()
```
mjeffw, line 1521: Does this need to go in the global namespace? I plan to reduce the use of the global namespace and rely on imports when that is possible.

templates/apply-damage/apply-damage-dialog.html
```diff
@@ -232,6 +232,7 @@ <h4>{{localize "GURPS.addSpecialCombatSituations"}}</h4>
                     <option value='0.5' {{select-if CALC.armorDivisor 0.5}}>(0.5)</option>
                     <option value='2' {{select-if CALC.armorDivisor 2}}>(2)</option>
                     <option value='3' {{select-if CALC.armorDivisor 3}}>(3)</option>
+                    <option value='3' {{select-if CALC.armorDivisor 4}}>(4)</option>
```
mjeffw:
```suggestion
                    <option value='4' {{select-if CALC.armorDivisor 4}}>(4)</option>
```

module/gurps.js
Legendsmith: Don't think so, I was just working as I saw CI implemented. However it may possibly need to do so if we implement alternate applydamage for CI.

PR #805, opened 2021-06-20
pass arguments to macros to use if Advanced Macros is installed

danielrab:
passing arguments to macros would allow users to basically make their own on-the-fly formulas.


danielrab:
I didn't realize there was such a thing as GURPS.chatargs & GURPS.chatreturn. This makes the change less relevant.

danielrab:
If the change to make Advanced Macros work nicely with GGA is made, them might as well support the return.

PR #833, opened 2021-06-25
Add Separate NPC Sheet for Conditional Injury

Legendsmith:
Move Conditional Injury NPC Sheet to its own Sheet
Update CSS for conditional injury NPC sheet
Fix incorrect misplaced double quote in npc-sheet.html too.

### DIFF
--- a/module/actor-sheet.js
+++ b/module/actor-sheet.js
@@ -1536,6 +1536,18 @@ export class GurpsActorNpcSheet extends GurpsActorSheet {
     GURPS.performAction(parsed.action, this.actor, ev)
   }
 }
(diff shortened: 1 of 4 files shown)


crnormand:
Quite possibly from Prettier Formatter.   We use this so that everyone's code follows the same pattern.   It will come in. behind the scenes and format older code so that it conforms.   It may be responsible... not certain through.

crnormand:
https://prettier.io/

Legendsmith:
Cool, I've installed that, the whitespace and all that should be fine now, I think this is good to go.

PR #834, opened 2021-06-25
Make Armor divisor code work with Survivable guns Special (4) AD Step.

Legendsmith:
This adds a (4) AD and does NOT slot it into the regular progression, instead it's treated as (3) with hardened by the code because it doesn't actually add a new step, it's just an alternate step. I don't think it needs to be put anywhere else.

### DIFF
--- a/module/damage/damagecalculator.js
+++ b/module/damage/damagecalculator.js
@@ -280,8 +280,9 @@ export class CompositeDamageCalculator {
     if (this._isExplosion) return 1
 
     if ((this._armorDivisor > 1 || this._armorDivisor === -1) && this._isHardenedDR) {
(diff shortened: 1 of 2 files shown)


crnormand:
I saw in your other pull request that you just added the '4' to the armorDivisorSteps.   Why did you do it differently this time?

Legendsmith:
Because I realized it's not actually a different step, just an alternate one that exists for Survivable Guns; as it adds an inherent (2) armor divisor to high velocity piercing weapons such as intermediate and full power rifles. In order for Amour Piercing ammunition to still function as normal, it uses (4) armor divisor; (3) would be too weak and (5) too strong. Thus it's essentially equivalent to (3) in that it's a step above (2) but below (5). However, it's NOT part of the regular steps; a (5) weapon attacking Hardened (1) DR doesn't drop to (4) DR, it drops to (3).

PR #925, opened 2021-08-18
more updates for TS

mjeffw:


mjeffw:
Notice: This PR and the previous one are work towards making our JS files work with Typescript types. Yes, we could have simply marked them as "don't analyze" but that does NOT move us towards TS adoption, really. The upshot of this is that we need to do as full of a regression test as possible. I'm working on that, but you have much more intimate knowledge about all the variations of OTF and Item functionality, so I need help in this effort.

I recommend that we get this testing done before doing anything else, as in the worse case scenario we may want to back out these two merges, which will become harder if we put more code changes on top.

PR #1043, opened 2021-11-07
Refactor perform action

danielrab:
Alright, so here's a summary of what I did:
each part of the `performAction` method was separated into a method inside the `actionFuncs` objects.
finding the best option for skills and attributes was offloaded to `findBestActionInChain`
the main job of `performAction` gets the best action from `findBestActionInChain` if needed and then performs the appropriate action

notes:
for now there's a method called `processSkillSpell`. Later I think it's best to make similar evaluation methods for other stuff and use them when evaluating, instead of returning early when running with `calcOnly`.
saving and restoring the modifiers bucket was refactored out of existence by performing all the calculations before any changes are made to the modifier bucket.
perhaps we can eventually make the methods that call `performAction` call methods in `actionFuncs` directly, thus removing the need for that function completely.


mjeffw:
I think this looks good! I'm going to merge it.

@danielrab - One of the next steps I'd like you to take is to move all the function definitions to its own file, and only leave the assignment to GURPS properties inside `gurps.js`. I want to clean up 'gurps.js` such that it only contains assignment of its properties, but no actual logic. (I.e., it contains the 'what' but not the 'how'.)

PR #1050, opened 2021-11-09
Added Stevil's color chooser code

crnormand:


mjeffw:
WE need to stop adding code in GURPS.js IMO. Add the functions elsewhere and add to the GURPS global object.

crnormand:
I agree... but as far as I saw, the only thing that was added to gurps.js was:

```
  ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
  // Added to color the rollable parts of the character sheet. Stevil...
  ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
  Hooks.on("renderSettingsConfig", (...args) => {
    addColorWheelsToSettings();
  })

  Hooks.on("renderActorSheet", (...args) => {
    colorGurpsActorSheet();
  })
```

As part of the init sequence, which currently resides in gurps.js.    Did I miss something?

mjeffw:
No. I didn't review the code. 

Even those hooks could be in the original file, though. As long as something calls them.

PR #1074, opened 2021-11-13
Updated Spells section.

mehalforc:
Updated "Foundry VTT.xml" to put in the missing RSL of Spells. Updated spells.hbs to add the title of "Pts" and use the missing RSL data for spells.


mehalforc:
Not sure why it is saying 3 commits. ddd1925 looks suspicious to me...
~Stevil

mjeffw:
It's just a merge commit.

PR #1077, opened 2021-11-13
Updated CSS for Spells and output for special case of 0 points for added Spell.

mehalforc:
Added CSS to make the Spells section of the character sheet fit better. Also changed the output for Spell level if a Spell was added at 0 points. Was displaying 'NaN' because imported value was 'null' for the Spell level. The imported GCS xml file has: '<level type="number">-</level>' for Spells added at 0 points. Quick fix for the display.

~Stevil


mehalforc:
Added the ability to read a CSS file, parse it, then look up values by using the selector and the attribute name. Then sets the default color values for the Color Picker. Was added so as the CSS Style Sheet changes for the selectors, there would be no need to update the Color Picker Defaults for color in 'miscellaneous-settings.js'.

Note:
This is using an Ajax call in synchronous mode. (async: false) = deprecated code.
Otherwise the file is not finished reading before setting the Color Picker default colors.
Will update it later when I find a better solution.

~Stevil

mehalforc:
Moving Color Picker part of program to own files.

mehalforc:
Rewrote parts and made some files eslint compatible...
~Stevil

mehalforc:
Fixed Issues...
Made code read css file without using synchronous reading of file.

PR #1085, opened 2021-11-19
Fixed the issue with the GGA loading slower.

mehalforc:
I removed the parsing call to load default colors by reading a CSS file. this was causing a slowdown for what was loading first. Other modules were loading before GGA was fully loaded in Foundry.


mehalforc:
Updated "color-character-sheet-settings.js" to remove the call and settings for the CSS parser. This was slowing down the loading of the GGA and other modules were loading before it, causing some errors in those modules. Put back the default settings for the colors.

PR #1139, opened 2021-12-03
added range/speed table

3Demon:
another range/speed table
1 -10 meters: 0
11 - 20 meters: -1
21 - 30 meters: -2


mjeffw:
I'd rather not integrate someone's house rules into the core system. Either this should be a module you can install along with the GGA or this PR needs to implement an open, customizable, range penalty system.

3Demon:
Maybe we can develop method for adding custom speed/range table?

mjeffw:
@crnormand Told me to accept this PR. (And I'm okay with it, honest!) 
;-)

PR #1143, opened 2021-12-04
Script to directly import .gcs character files

rinickolous:
This relies on some data which is not saved by GCS in its current version (v4.35.0).
Should be fine to pull after the new version of GCS, which includes the needed data saving, is released.
Users will have to open their character files in GCS, modify them in any way, then Save them again, for the needed data to be included in the file.

Next step is mass import, as requested by Ben#4565 on Discord.
I didn't do much in terms of error-proofing, assuming users are not manually modifying their GCS files.


rinickolous:
I'm noticing now that some of the commits are unrelated (the Squashed commit, and branch merge). I hope that doesn't cause any issues.

rinickolous:
I'm also working on importing whole folders of characters. Should get it done before Monday evening UTC.

rinickolous:
There are a few bugs I noticed during play that I need to fix.

rinickolous:
Note: There are still a few issues I need to iron out with this. Namely, SM and Move don't seem to import properly.

crnormand:
No worries... we don't have to publish the fact to the world, just yet.    Plus this allows Nick and I to play with it also (since we only use GCS anyway).  Sorry it took so long to merge.

PR #1255, opened 2022-01-23
Added support for Unmodified Damage in the GCS and GCA4/5 files for use with GGA import.

mehalforc:
Updated the import files for GCS and GCA4/5. It supports a new tag called "unmodifiedDamage" that is exported in the XML, just under the "damage" tag for the melee and ranged attacks, in the XML.


crnormand:
Ok, why do we want this?   (I know we do... just not why)

mjeffw:
So we can adjust Damage when active effects as ST.

PR #1455, opened 2022-07-06
make PDF links respect offset setting of PDFoundry

danielrab:


crnormand:
PDFoundry internally adds the offset.    By adding it again, we are getting double offset.  At least that is what happened when I included this code.

Do you have a testcase where it isn't respecting the PDF offset?

danielrab:
that's weird, I could swear it was not doing it when I tested it before, but now that I test it again, it does indeed add the offset.

PR #1545, opened 2022-09-09
Typescript Reorganization

rinickolous:
Reorganization of files based on default [foundry-factory](https://www.npmjs.com/package/@ghost-fvtt/foundry-factory) setup.
There is currently an issue wherein `lib/xregexp-all.js` is not being read properly.
This also replaces (nearly) all references to "X.data.data." with "X.system", or at least what I could find with grep.
The automatic formatting differs from the currently used one in the case of semicolon use, default quotation marks, and indentation style. This can be discussed/adjusted before merging.


rinickolous:
On second thought, it may be more worthwhile to do the v10 implementation in JavaScript, and switch over to TypeScript only for the merge. Will being working on that now

rinickolous:
Closing this for now

PR #1841, opened 2023-08-10
feat: adding a reducedMode character sheet view

Azim-Palmer:
image

- I've taken as light a touch approach as possible. I've modified one existing hbs template (adding a class) to better target it with css selectors.
- I've put the styling `reducedMode` overrides next to their overridden elements, would be happy to move them to one block.


rinickolous (approved):
Looks good. Merging.

PR #1941, opened 2024-08-03
Fixing issue with import of version 5 equipment lists

jmcconn18:
The "type" property has been removed on V5 equipment lists, so we can no longer validate on that. Also, the "type" property has been removed from weapon attacks, so those have been changed to a check for existence of property "reach" for melee attacks, and "range" for ranged attacks. Should be backwards compatible with older versions, but I saw no obvious way to check to see if an object is an equipment list in V5.

### DIFF
--- a/module/item-import.js
+++ b/module/item-import.js
@@ -17,10 +17,8 @@ export class ItemImporter {
     } catch {
       return ui.notifications.error('The file you uploaded was not of the right format!')
     }
-    if (j.type !== 'equipment_list') {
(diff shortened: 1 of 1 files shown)


mjeffw:
I independently fixed this and my changes are already merged. I'm going to close this PR without merging. I truly appreciate your help and I only wish I had seen this PR before! Thank you.

jmcconn18:
> I independently fixed this and my changes are already merged. I'm going to close this PR without merging. I truly appreciate your help and I only wish I had seen this PR before! Thank you.

No worries, thanks for all the work you do on the project!

PR #1943, opened 2024-08-08
Fix random horizontal scrollbar in character sheet

Wilou1428:
Currently, there is a useless scrollbar in the character sheet window. 
image

I don't think this does anything useful. Seems like a leftover style sheet rule. I removed it, seems to have fixed the issue.

### DIFF
--- a/styles/simple.css
+++ b/styles/simple.css
@@ -518,7 +518,6 @@
   padding: 5px;
   scroll-behavior: smooth;
   background: white;
-  overflow-x: scroll;
 }
 
 .gurps .sheet-header {
--- a/styles/simple.css
+++ b/styles/simple.css
@@ -518,6 +518,7 @@
   padding: 5px;
   scroll-behavior: smooth;
   background: white;
+  overflow-x: auto;
 }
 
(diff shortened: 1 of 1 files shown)


mjeffw:
I actually just added that CSS style to the last release. It's needed for folks who scale the display because of vision impairment. We won't remove the scroll bar but I'll look into how to prevent it from appearing if the window is large enough to display the whole sheet width.

Wilou1428:
Oh I see. I can try looking into that myself.

Wilou1428:
Is there a way to for me to set up the screen so I can test it? Because zooming in the popped out character sheet I don't see the difference.

Wilou1428:
I found out what was causing this. I noticed that the bar only appeared when the character sheet had notes. Apparently, the note elements extend further than it's visible causing the bar to appear. So my solution would be the to hide overflow for the notes section, this shouldn't impact the Pop-Out large image case. Also set the `overflow-x: auto` so it would hide the scrollbar if there's no overflow.

mjeffw:
Looks good! Thanks for the help.

PR #1958, opened 2024-08-22
Feature/ht hitlocations

NoOrangeSkitty:


mjeffw:
First -- thanks! Right now I'm the only developer of the original three that has time to work on this project, and your help is appreciated.

I'm about to review the changes and if there are no issues, I'll merge this PR.

One point about process: The standard process is to branch from main, and create a PR to push back to main. You've chosen a "work in progress" branch for my Tabbed Sheet updates. That's okay right now because I just finished the work on that branch and can look at your PR. But it could be problematic to create a PR on top of a WIP branch; in some extreme cases, the WIP branch may be deleted without merging (a failed experiment, for example, or an abandoned feature).

So in the future be sure to branch from main and push to main.

NoOrangeSkitty:
Sorry to hear that, I knew one of you left but not two. I would love to keep helping as I currently have quite a bit of free time and need more programming experience. Also, after the merge I caught a mistake I made, when programming in the roll to determine the location I copy and pasted the random hit location code but accidentally left in some comments, it should be right above the random hit location code, that function could also be slightly optimized.

PR #1959, opened 2024-08-23
Use `ruler._addWaypoint` instead of modifying waypoints directly

### LINKED ISSUE #1948: Weird interaction between Effects Modifier and Elevation Ruler module
Not sure what causes this (or if it's replicatable), so posting both here and in the Elevation Ruler github:

AFAICT, when Elevation Ruler is enabled, the Range calculation for the Effects Modifier has 4 units added to it. I've tried with both "y" and "m" as the "Units" value, and in both cases a Foundry ruler distance of 10 becomes 14 and a distance of 22 becomes 26 in the Effects Modifier. This seems to be a ER issue, but as this functionality is relatively recent, and in that the miscalculation does not affect other measurements involving ER (at least at my end) it might be a bit more complicated than that.

caewok:
For better compatibility with Ruler updates and modules that may modify the Ruler, use the `_addWaypoint` method instead of directly modifying waypoints.

Note that `calculateRange` may be incorrect for tokens with differing elevations. Default Foundry v12 does not handle this situation although mods like Elevation Ruler can. Note also that `canvas.grid.measurePath` might be the better approach instead of constructing an entire Ruler for one measurement.

Closes #1948.

### DIFF
--- a/module/actor/effect-modifier-popout.js
+++ b/module/actor/effect-modifier-popout.js
@@ -82,7 +82,7 @@ export class EffectModifierPopout extends Application {
     // const ruler = new Ruler() as Ruler & { totalDistance: number }
     const ruler = new RulerGURPS(game.user)
     ruler._state = Ruler.STATES.MEASURING
-    ruler.waypoints = [{ x: token1.x, y: token1.y }]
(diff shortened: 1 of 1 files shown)


mjeffw:
Thanks!

PR #1968, opened 2024-09-02
Fixed Body Hits not capping torso/groin damage

NoOrangeSkitty:
I noticed that torso and groin damage was not being capped when use body hits was enabled and found that the _usebodyhits variable was undefined in the pointstoapply function in damagecalculator therfore the if statement that capped damage would never trigger, i fixed it by defining _usebodyhits at the start of pointstoapply.


mjeffw:
Already fixed. Thanks, though!

PR #2017, opened 2024-10-31
Release 4

### LINKED ISSUE #2007: Damage dropping not working as intended
When I drag damage to a hex, it should apply damage to that character. I understand if there is more than one character in the hex. But there isn't. And it asks who I want to apply it to. Here, that's not so bad--it's easy enough to see the difference. But when you have five guys named "Guard" it can be maddening. Yeah, the workaround is to give everyone different names, but the problem is it shouldn't be doing that.

[discord link with video example](https://discord.com/channels/228546929494065152/788878698349723669/1298106229482328084)

### LINKED ISSUE #2010: [Bug] The "roll for me" option does not work
The "roll for me " option when using the Physical Dice setting seems to have broken for me - it returns only zero as result.

2024 10 20 - 22h10m28s

### LINKED ISSUE #2011: Add a setting to toggle  system wide "if Attack roll succeeds, roll damage automatically"
I know this can be done via On the Fly formulas, but I think this would be handy as a system wide option - especially for the GM. 

As side thought, something like: "if Token A has a target, and Token A succeeds an Attack roll, roll damage for the Attack and Dodge for the target", could be interesting too. But this has a narrower use case.

### LINKED ISSUE #2013: [Feature Request] Improving the combat system
Good afternoon. Is there an option in the system to add default macros for an attack and be able to respond to this attack (for example, a button with defense in a chat)?
An example of how this is implemented in a different system for gurps:

https://github.com/user-attachments/assets/37241949-9f57-4ca8-8e7a-a89d401fd96f

### LINKED ISSUE #2027: Request: Maneuver Menu on Combat Tracker
This feature is part of the list of improvements proposed in #2013.

The purpose is to make it easier for the GM to manage each combatant's maneuvers in the Foundry's Combat Tracker.

It's simple to use, just click on the button to select a maneuver for the combatant. To remove the selected maneuver, just right-click.

This button is visible only after that token rolled for initiative. The initiative number is now in the Token's image tooltip.

GGAManeuverButton

The tooltip gives tips about the selected maneuver:
* The token's maximum movement - calculated from current move vs the selected maneuver
* Whether the token can attack
* Whether the token can defend

### LINKED ISSUE #2028: Request: Quick Roll Menu
This feature is part of the list of improvements proposed in #2013.

The purpose is to make it easier for the GM to handle each combatant's rolls in the Foundry Combat Tracker.

To use, just click the button to open the menu with the quick rolls selected for that actor. When opened for the first time, you will see this:

image

* **Checks**: the secondary checks for the Actor, including base damage.
* **Defenses**: the selected defenses for the Actor. On default configuration, only Dodge is selected.
* **Attributes**: The main attributes checks for the Actor, including Perception and Will rolls.

These options can be configured in Settings: 
image

image

### Making blind rolls
Click on the Eye icon on the menu to toggle between regular and blind rolls. For GMs, the default behavior is blind rolls.

### Adding new Rolls

To add new rolls to Quick Roll, open the Actor sheet from the Token or from the Actor list, and on the correspondent Actor Component (or if you're using Foundry Items, the Actor Item) select the `Add to Quick Roll` option for that component/item. 

Please note:
* When opening the Sheet from the **Token**, you will change only the Quick Rolls for _that_ token.
* When opening the Sheet from the **Actor**, you will change the Quick Rolls for **all** tokens _linked_ to that actor.

GGAQuickRoll

When you add a melee attack to Quick Roll, system will add:
* The Melee Attack
* The Block and Parry for that Attack, if they exist.

Also, for Melee and Ranged attacks, hold `crtl` key before click to roll for damage instead the attack roll:

GGAQuickRollDamage

### Adding OTF
Just add the Component or the Item to the Quick Roll. System will search for OTFs on the `Notes` field for skills, spells and traits:

GGAQuickRollOTF

### LINKED ISSUE #2029: Request: Tagged Effect Modifiers
Part of improvements suggested on #2013 and #1969

The purpose of this improvement is to add a decision-making logic to the system so that it can automatically add modifiers before rolls.

## How do tags work?
Tags are identifications that allow you to cross-reference the "sources" of the rolls with the existing "modifiers":
* In the _effect modifier_, you add the tags that identify that bonus or penalty. Ex. `+1 to camouflage #hide`
* In the skill, spell, attack, defense or damage (the _source of the roll_), you tell the system which tags it should look for before making the roll. Ex. `hide, scout`

## Where modifiers can be saved?

Usually a third party is the reason for the existence of this modifier, or in other words, the _modifier's origin_. For example, it can be an advantage, a piece of equipment, a spell, etc. This is where you can optionally save the modifier.

You don't need to find the modifier's origin - it's just a way to organize them. In fact, you can add all your modifiers in Character Sheet's Conditional and/or Reactions sections or inside the Effect Modifiers Popout.

With this in mind, the options for saving modifiers are:
* In the trait that originated the modifier, within the Actor Sheet.
* In the `Conditional Modifiers` section of the Actor's Sheet
* In the `Reaction Modifiers` section of the Actor's Sheet
* In the `Effect Modifiers` popup of the system.

## Modifier with tags format
The format is `<modifier> <description> <#tag1> <#tag2>...` 
* You can add as many tags as you want. 
* Tags accept only letters, numbers and underscores. `#foo` is good, `#b@r` is bad. 

## Example using Foundry Items:
Suppose your Cloth armor (the equipment, or the modifier's origin) gives +3 to all camouflage tests. Opening the armor item, and navigating to "Effects" you can add the armor modifiers here:

image

In the `Effect Modifiers` popup you will see the modifier created:

image

Now, going to the "**Camouflage**" skill we add the tags that it should look for before rolling.

image

When requesting a skill roll, the system will search for all modifiers with the tags configured in the skill and add them to the player's Bucket:

GGATaggedRoll

If you have the `Show Roll Confirmation Dialog` setting enabled, you can check your bucket before rolling. Otherwise, the roll will be made with the selected modifiers.

## Example with Actor Components
Suppose you house ruled the "Intuition" trait gives +1 to the "Meditate" skill. On Intuition, you can add the modifier:

image

Then, add the tags the skill "Meditate" will lookup:

image

## The `Effect Modifiers` popup
Use this window to create custom modifiers and to see the list of Actor modifiers. **Only the modifiers listed here** will be searched for the roll and be added automatically if they have the correct tags.

image

This window shows the following information:
* The modifier's description.
* The modifier's origin or the source of the roll (calculated by the system).
* The modifier's tag list.

Modifiers with a "**custom**" origin are those created by the user. 
* To create a new custom modifier, use the (+) button.
* To update the list, use the **refresh** button.

### Deleting modifiers
* To delete a custom modifier, simply right-click on it in the popup.
* To delete a saved modifier, on right-click, system will open the Item or the Sheet (for actor components)
* To delete all modifiers, use the delete button. Note that if you click the `refresh` button, all modifiers saved on the player's sheet will be added again.

## Tags for rolls without items
For rolls that do not have an item or actor component to represent, such as attribute tests (ST, Will, etc.), character tests (vision, fright check, etc.), you can define (or delete) the tags that should be searched for these rolls, in the Foundry settings:

image

* The same configuration screen applies to configuring or deleting "groups" of rolls. Example: all rolls, all attacks, all spells, etc.
* On this screen, you can also disable automatic calculation in the Bucket, if you wish.

## About Damage Rolls
This works for damage rolls too, and for large modifiers - ex.: 1d+10 - system will convert the modifier to additional dice, as per GURPS optional rule (B377)

GGATaggedDamage

### LINKED ISSUE #2030: [Feature] Request: Confirmation Dialog Roll
This feature is part of the list of improvements proposed in #2013 

The purpose of this feature is let the GM or the player review their modifiers before the roll.

## Enable or disable the dialog
Use the settings `Show Confirmation Roll Dialog` - the default is enabled.

image

## Usage
Usage is very simple - just click on a roll:

GGAConfirmRoll

All changes made in Bucket will be reflected on the dialog. The colors on the total roll is a hint of the roll probability of success.

This dialog works for:
* All checks
* Damage Rolls
* On-The-Fly Formulads

### Whats happens if no Actor is selected?
This can happens in some cases, such as the 3d6 button. For these cases, the system will roll without show the dialog.

### LINKED ISSUE #2031: [Feature] Request: Check Maneuver, Max Actions and Targets before Roll
This feature is part of the list of improvements proposed in [https://github.com/crnormand/gurps/issues/2013](https://github.com/crnormand/gurps/issues/2013).

## What will the system do?
- For **Attack** and **Defense** rolls, the system will check if the selected token's maneuver is valid for that roll.
- For **Attack** and **Spell** rolls, the system will check if the user has selected targets using Foundry's targeting system.
- For **Attacks**, **Blocks**, and **Spells**, check if the user already has reach the maximum number of actions for that round.
- These checks only occur when combat is active and the token is a combatant.

## Usage
Define in the settings what the system will do if it finds an invalid situation:

image

The available options are:
- **Allow**: Just allow the action, no questions asked.
- **Warn**: Warn the user in the Confirmation Roll Dialog. If the dialog is disabled, no warning will be shown.
- **Forbid**: Prevent the roll and alert the user.

The default value for both checks is `Warn`.

Then, on Character Sheet, define the maximum number of actions this character have per combat turn (or foundry round):

image

* `Max Actions per round` is the max number of actions the player can do in that round, excluding skill and dodge rolls.
* `Max Blocks per round` is the max number of blocks the player can do in that round.

Some maneuvers can add a additional action for the turn (ie. AoA Double). 

### Warning

Warning Message

### Forbid

Forbid Message

### LINKED ISSUE #2033: Request: Add Maneuver's and Parry modifiers automatically before roll
This feature is part of the list of improvements proposed in #2013

## Maneuvers
System will include, on players `Effect Modifiers Popout`, the modifiers based on the following maneuvers:

* **Aim**: the ranged bonus, after the first round aiming, for all ranged weapons.
* **Evaluate**: the bonus, after the first round, to any attack roll
* **All out Attacks**: the bonus to attack (varies)
* **All out Defense**: the bonus to defense (varies)
* **Parry penalities**: the cumulative penality for consecutive parries, using the same weapon, on the same round.

### Example
Bree MacFadden is fighting two cultists. On first round, she acts first, and choose AoD Parry:
image

The first cultist hit his attack. Bree choose to parry the attack:
image

The second cultist hits too. She decides to parry again, and because she is a Weapon Master, their next parry is at -2:
image

### Setup Parry base penalty.
On character sheet, open the melee attack to define the base Parry penality. Default is -4, but you can define any number here, to support _Weapon Master_ or _Trained by a Master_ traits.

image

### LINKED ISSUE #2034: Request: Improved support on Injury effects on Apply Damage Dialog
This feature is part of the list of improvements proposed in #2013

On ADD, system calculates and return the target effects based on injury to be applied. This feature adds new funcionalites:

* **"Save Roll" button**. For the effects with need to check HT before apply.
* **Add Shock to target's next turn button**. System will add the calculated shock effect from injury on the start of the target's next turn.

To remove effects and the scheduled shock effect, just click on the button again.

### Example
GGAAdd Shock

chrismaille:
### Features
* Close #2027
* Close #2028
* Close #2029
* Close #2030
* Close #2031
* Close #2033 
* Close #2034

### Bugfixes
* Close #2007
* Close #2010 

Can close #2013, can close #2011


mjeffw:
@chrismaille 
Confirmation dialog: The actual numbers seem somewhat hit or miss. See the following:
image

This is the result when I clicked on "Arrow (Fine)" in the Composite Bow attack. The Effect Modifier popup shows (correctly) -1 range to target Enok. But both the bucket and the dialog reflect a -2 modifier for range. Further, the skill is shown as "UNDEFINED-19".

Without changing anything, if I then click on "Bodkin" usage of the Composite Bow, the bucket now shows "+0" but the dialog shows -1 modifier.

image

mjeffw:
[fvtt-Actor-bjorn-bjornsson-Qbyxek8wckIgivNT.json](https://github.com/user-attachments/files/17961211/fvtt-Actor-bjorn-bjornsson-Qbyxek8wckIgivNT.json)

Here's an export of the "Björn Björnsson" character.

mjeffw:
@chrismaille 
When using Foundry items, is there a way to turn off the image on the character sheet? 
image

chrismaille:
> @chrismaille When using Foundry items, is there a way to turn off the image on the character sheet? image
> 
> Maybe even per trait type? I could see the utility or aesthetics of showing the image on equipment, but Advantages and Skills may not have the same utility.

Settings are here:
image

You can toogle by sheet section, but not on a item individually

image

chrismaille:
> [fvtt-Actor-bjorn-bjornsson-Qbyxek8wckIgivNT.json](https://github.com/user-attachments/files/17961211/fvtt-Actor-bjorn-bjornsson-Qbyxek8wckIgivNT.json)
> 
> Here's an export of the "Björn Björnsson" character.

The JSON file did not imported the ranged attacks for me - can you send to me the GCS/GCA exported file?

mjeffw:
[Bjorn Bjornson-20221104.gcs.zip](https://github.com/user-attachments/files/17962776/Bjorn.Bjornson-20221104.gcs.zip)

chrismaille:
Ready for tests

mjeffw:
@chrismaille 
In the ADD, Knockback is not being handled  appropriately. You reference `object.modifier` in the code below, but that value is undefined.

```js
  async _handleTestSaveEffectButtonClick(ev) {
    let dataEffect = ev.currentTarget.attributes['data-effect'].value
    let effect = JSON.parse(dataEffect)
    let otf = ''
    switch (effect.type) {
      case 'headvitalshit':
      case 'majorwound':
        const htCheck =
          effect.modifier === 0 ? 'HT' : effect.modifier < 0 ? `HT+${-effect.modifier}` : `HT-${effect.modifier}`
        otf = `/r [!${htCheck}]`
        break
      case 'knockback':
        const dx = i18n('GURPS.attributesDX')
        const dxCheck = effect.modifier === 0 ? dx : `${dx}-${object.modifier}`
        const acro = i18n('GURPS.skillAcrobatics')
        const acroCheck = effect.modifier === 0 ? acro : `${acro}-${object.modifier}`
        const judo = i18n('GURPS.skillJudo')
        const judoCheck = effect.modifier === 0 ? judo : `${judo}-${object.modifier}`
        otf = `/r [!${dxCheck}|!SK:Acrobatics|!Sk:${acroCheck}|!SK:Judo|!Sk:${judoCheck}]`
        break
    }
    if (!!otf) await this.actor.runOTF(otf)
  }
```

mjeffw:
Another issue -- when applying an effect that changes the posture, it appears the code is "toggling" instead of setting the new effect. For example, if a character is prone and they take a major wound, applying the major wound effect (prone + stun) makes him stand up.

chrismaille:
### Fixed Knockback and effect toggling on ADD:
GGAAddEffects

### Shock at next target turn
This is the standard rule (as per B381 and B419), but point taken, we don't need to enforce this. Use the new settings to define apply shock at next turn or immediately:
image

### Effect Modifiers Popout Size
I did not exactly reduce the size (not sure how yet, there is a lot of information on this window now) - instead I define a max height for the Modifiers main list:
GGAScrollEffects

Ready for tests again.

mjeffw:
> This is the standard rule (as per B381 and B419), but point taken, we don't need to enforce this.

The rules if read literally support your interpretation, but it leads to silly effects, such as a character is hit and before his next turn has to make a IQ roll to see a spell being cast by another combatant; as written this roll would not suffer the penalty. As soon as it is his turn he suddenly has these penalties. One can argue that the roll to see the spell being cast happens on his turn, bit that's not how most people sequence the rolls.

mjeffw:
> Effect Modifiers Popout Size

Can individual conditional modifiers be deleted out of the list?

chrismaille:
> The rules if read literally support your interpretation, but it leads to silly effects, such as a character is hit and before his next turn has to make a IQ roll to see a spell being cast by another combatant; as written this roll would not suffer the penalty. As soon as it is his turn he suddenly has these penalties. One can argue that the roll to see the spell being cast happens on his turn, bit that's not how most people sequence the rolls.

This is a very old question, from 3rd edition I think. I remember Kromm said this is because every character's turn occurs at the same time.

> Can individual conditional modifiers be deleted out of the list?

It's a possibility. My concern is the players start to flag false bugs, because they are not seeing all the modifiers available. I was thinking to add a filter on this window, but my suggestion is wait to see how players will adapt to tags first.

Issue #2039, opened 2024-11-28
Bug: Lifting and moving things chart breaks with high ST characters


PrinceOfIonia:
Exactly as the title says. I am running Version 12 Stable Build 331 of FoundryVTT.

This character has 100 ST, this is what GCS shows: 
![image](https://github.com/user-attachments/assets/2a16b9e7-b4db-4f56-b0f8-389b3f0931f7)

This is what GGA shows when the sheet is imported:
![image](https://github.com/user-attachments/assets/d496265b-bbeb-43a1-9423-35f9d2541214)

Here is the GCS file: 
[GCS File.zip](https://github.com/user-attachments/files/17952824/GCS.File.zip)

Let me know if you need more info or have any questions.

PR #2041, opened 2024-11-29
Fix bl unit for high strength characters

StefanLeng:
GGA 0.17.16 fixed issue #2039 for the Basic Lift value, but the unit is still broken. Its not displayed on the GGA sheets, but I use it in the Cathegorized Character Sheet module.

This pull request extends the fix to also fix the unit.

### DIFF
--- a/module/actor/actor-importer.js
+++ b/module/actor/actor-importer.js
@@ -1537,8 +1537,9 @@ export class ActorImporter {
     }
     data.QP.value = qp
 
-    let bl_value = parseDecimalNumber(calc?.basic_lift.match(/[\d,\.]+/g)[0])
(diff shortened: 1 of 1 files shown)


mjeffw:
@StefanLeng Do you want it displayed such as "2000 lb"?

StefanLeng:
Yes. I just display the strings from system.liftingmoving, which includes the unit. Works for "normal" characters, but if the BL >= 1000, this string is still broken in 0.17.16. With a basic lift off 2000, it was "2 2,000 lb" in 0.17.15 and now is "2000 2.000 lb" in 0.17.16. This pull request makes it "2000 lb".

See https://github.com/StefanLeng/gurps-categorized-sheet/issues/9 for the situation under 0.17.15. Its still broken under 1.17.16, even if the leading value is now fixed.

PR #2114, opened 2025-01-11
2112 1717 b4 roll confirmation breaks multiple damage rolls even if disabled

### LINKED ISSUE #2112: 17.17-b4 Roll Confirmation breaks multiple damage rolls even if disabled
This works if Roll Confirmation is disabled; if enabled it only rolls the last one: `/r [2d (2) pi] \\/r [1d burn]`

The issue happens even with separate executeOTF statements. 

```js
let attackDamageOtF = '[1d-4 cr]'; // [6d+14 imp] or [2d+4 cr] or [2d+4 (2) cr] 
let followUpDamage = '[1d-4 burn]'; // default follow-up damage

function wait(x) {
    return new Promise((resolve) => {
      setTimeout(() => {
        resolve(x);
      }, 10000);
    });
  }

await GURPS.executeOTF(`/r ${attackDamageOtF}`); 
await wait(1);
await GURPS.executeOTF(`/r ${followUpDamage}`); 
```
With Roll Confirmation enabled, the added delay lets both rolls work as long as you click the first Roll Confirmation fast enough. However, if you disable Roll Confirmation, the second executeOTF never runs.

Using regular chat rolls, this next OtF works only if Roll Confirmation is enabled; if not enabled it only rolls the last one:
`["Explosion with Fragments" /4d-1 cr\\/2d cut]`

mjeffw:
This fixes #2112 but I worry about code paths (such as exceptions) that could leave the semaphore locked.

### DIFF
--- a/module/gurps.js
+++ b/module/gurps.js
@@ -678,7 +678,7 @@ if (!globalThis.GURPS) {
           await dialog.render(true)
         }
 
-        openDialogWithSemaphore(dialog)
+        await openDialogWithSemaphore(dialog)
         // await $(document).find('.dialog-button.cancel').click().promise()
         // await dialog.render(true)
       } else {
--- a/module/gurps.js
+++ b/module/gurps.js
@@ -84,6 +84,7 @@ import { AddMultipleImportButton } from './actor/multiple-import-app.js'
 import { TokenActions } from './token-actions.js'
 import { addQuickRollButton, addQuickRollListeners } from './combat-tracker/quick-roll-menu.js'
 import { addManeuverListeners, addManeuverMenu } from './combat-tracker/maneuver-menu.js'
+import Semaphore from '../lib/semaphore.js'
 
 let GURPS = undefined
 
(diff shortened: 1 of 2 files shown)


module/gurps.js
```diff
@@ -618,56 +621,60 @@ if (!globalThis.GURPS) {
...
-          render: html => {
-            html.closest('.window-app').css('height', 'auto')
-          },
-        })
         // Before open a new dialog, we need to make sure
         // all other dialogs are closed, because bucket must be reset
         // before we start a new roll
         await $(document).find('.dialog-button.cancel').click().promise()
-        await dialog.render(true)
+        await new Promise(async resolve => {
```
chrismaille, line 628: @mjeffw Main issue was here: this dialog must be inside an awaitable promise, because multiple line OTFs call this dialog inside a forloop on chat.js `ChatProcessorRegistry.processLines()`

chrismaille:
Ready for tests

PR #2200, opened 2025-04-20
Started on GurpsActorSheetV2

rinickolous:
just some preliminary work, setting up files to fill in later.
This changes also includes some foundational work for working with .ts files alongside .js files.


rinickolous:
@mjeffw I want to run this first one by you before merging as it does include some foundational changes which I want to make sure I've gotten right.

mjeffw:
Give me a breakdown of each change, please.

rinickolous:
# Major Changes

## Introduction of typescript & tsup

The way I currently have this set up is tsup is processing any .ts files and emitting them in-place. This may not be ideal, as it requires that the build process is run at least once so that the changes carry over to the .js files. It also effectively produces duplicate files for each .ts file, which isn't ideal either.

The alternative to this approach would be to split up the project into src/ and dist/ folders and modify the build process to deploy the ready .js files (or a bundle thereof) into the dist/ folder, which is .gitignored. This would also require changing the GitHub release automatic action to act accordingly when creating releases. I chose not to do this (for now) as it would be immediately disruptive to dev workflow (you'd have to re-do symbolic links, it would create a dev incompatibility with other potentially active branches due to the large amount of rename operations and potentially conflicting commits, etc.). As the included .ts files are small in number, I went with the first approach for now, but would ideally want to switch to this one later.

I can instead go with the latter approach now, but want explicit sign-off for that from you beforehand because of how disruptive the change would be.

## Alternative Handlebars Partial Handling

As I mentioned over discord earlier, I think it would be a long-term benefit to re-do our handlebars partial handling so that the keys corresponding to partials include more contextual content (i.e. folder structure in the filename) `utils/handlebars-partials.js` includes this alternative implementation, which would (for now) live alongside the current implementation. Due to the nature of both, there would be no real possibility of conflicts.

### Comparison between new handling and old
- Under old handling method, file with path "systems/gurps/templates/actor/parts/header.hbs" would use key "header" (taking into account only filename minus extension)
- Under new handling, the same file would use key "actor.parts.header", which is more informative and less likely to produce conflicts if multiple similar parts are needed

I want to introduce this and use this alongside the current handling to avoid unnecessary refactoring which would effectively be system-wide otherwise.

## Key separation in language files

I started using non-flat references in the language files when referring to elements which appear in re-done applications. I think moving to this sort of structure for the files long-term would be beneficial for the sake of better categorising and organising localizable strings.

This change is disruptive in that for full introduction the existing localization files would have to be amended so that these elements are included. If needed, I can recreate the same JSON structure for our existing translations before merging this PR and submitting additional PRs, to ensure localization isn't lost. I can also undo this change and stick with the existing flat keys for now

To make the best use of this (and minimize refactoring later), I believe we should have a discussion about how to break down this structure, and produce a standards document to stick to.

## Sheet tag and CSS changes

While this process has not started properly yet, I want to use migration to AppV2 as an opportunity to reconsider some of the structure and styling choices made for our applications historically. This includes
- More prominent use of semantic tags rather than divs: This moves us towards better style consistency with the rest of the FVTT application and improves accessibility.
- Separate CSS style definitions for new applications: Introduce a clean break in terms of CSS classes etc. used to define styles, to better adhere to FoundryVTT's existing styling and avoid some of the potentially overactive or poorly defined styles which may have slipped in during development historically.
- (not implemented yet) Introduction of LESS/SASS: I propose introducing LESS/SASS (ideally SASS, but it's mostly just a preference) for better organisation of our CSS styles. The proposed implementation would be done in such a way that the current style definitions are not disrupted. This would, again, warrant a discussion about standards for file/folder breakdown for CSS styles, and would (again) introduce a build step into the dev workflow.

mjeffw:
We should start adhering to proper git usage. Commits and PRs should be small to allow for speedy review and resolution. Each logical change should be isolated in its own PR. You've listed 4 big changes above; this should be 4 different PRs.

### Key separation in language files
Regarding the lang file restructuring, I am all for the approach you are advocating. In fact I spent a couple of weeks very recently doing the same thing, only to discover that this change cannot be made without coordinating changes with other projects, such as StefanLeng's GURPS Categorized Character Sheet, the Nordlond Bestiaries, and Token Action HUD. So I had to undo the changes.

Some keys should be "universal" in nature, and not tied to an application. Attribute names (ST, DX, etc) and common language stuff (Cancel, OK, Submit, etc.) are examples. I suspect most tags fit that description and what is unique to an application would be phrases such as the settings descriptions.

Also, if we touch one lang file, I want the same change made to all languages files so that they stay in synch.

### Alternative Handlebars Partial Handling
I agree with this, submit it as a standalone PR request for my review.

### Introduction of typescript & tsup
I understand the approach where there is a dist/ folder, that's the one that GCSGA used. In the proposed approach, I assume that you'd have some files like `module/actor/actor-sheet.ts` and it would generate `module/actor/actor-sheet.js`. Correct? Is that the only extra file created? You mentioned "It also effectively produces duplicate files for each .ts file" so I want to be sure I understand the change.

Would those .js files be committed to.git? Or should we add them to the .gitignore? I don't like the idea of committing generated files into the repository.

Please submit this as a separate PR.

### Sheet tag and CSS changes
Again I mostly agree with the proposed changes. Give me a separate PR for review.

PR #2202, opened 2025-04-22
Adding foundational support for Typescript

rinickolous:
This change adds the foundations for future Typescript support / migration, while making effectively no content changes to the code.

# Breakdown of changes
### rollup (+ plugins)
This dependency bundles the GGA code from the src folder and puts the result in the dist folder. This gets around an issue in a previous commit which would otherwise commit us to including generated files in the GitHub codebase. There is an important caveat with this: Our current GitHub actions **will not work** with this as they stand now. I can add those to this PR, but am submitting it for review now before doing so. The required change would basically consist of adding a build step to the GitHub actions to produce a built version of the package (because there's now a build step)

### Autoprefixer
Just a QoL plugin which will automatically add prefixes to CSS variables which use experimental or browser-specific features. Not in use anywhere yet but will come in handy.

### fvtt-types
The most up-to-date and reliable type system for FoundryVTT as things stand now. Currently using the documents-v2 branch to remain on the cutting edge, but subject to change.

### prettier
I'm not sure how this was not a dependency already if we had a prettier config defined. Have we been relying on a version included in an IDE?

### sass
To be used for extending CSS functionality with things like modularity and nesting. Not in practical use yet but good to add now in my opinion.

### typescript
Self-explanatory, I hope.

## Moving files
As mentioned above, all source files have been moved to src/* where * is their original position.
This has been done with the exception of gurps.js, which as been moved to src/gurps.js rather than src/module/gurps.js. The related configs also reflect this. I'm not married to the change concerning gurps.js specifically, but it feels like the right thing to do.

## Defining typescript paths
Currently, a single "@/*" typescript path is defined, but I have no intention of keeping it like this. I think the use of only relative imports throughout the codebase is rather messy and would like to swap this out for well defined and sectioned off paths in the future. This just lays a foundationw without making any huge sweeping changes.

<hr>

This PR is not ready for merge, as it is missing the necessary workflow files needed to not break the release/pre-release process. Please review the current state and let me know if anything needs changing at this stage.

I believe the benefits of moving to this basically speak for themselves as far as long-term support and developer experience goes.


src/templates/actor/tracker-manager.html
mjeffw, line 1: When moving/renaming these templates, go ahead and rename the extension to .hbs for consistency. Same comment for all *.html files in this directory.

test/parselink.test.js
```diff
@@ -1,12 +1,19 @@
...
-}
+import { parseForRollOrDamage, parselink } from '@/lib/parselink.js'
+import { DamageTable } from '@/module/damage/damage-tables.js'
+
+beforeAll(() => {
+  /** @type {Game} */
+  globalThis.GURPS = {}
+
+  // HACK: this is used to get around testing in a non-Foundry environment
+  // it should be removed in favour of a proper test environment
```
mjeffw: This isn't a hack. It's a necessary part of running a unit test outside of a node container or browser. Unless you know something I don't -- let me know if so. Otherwise remove the comment.

mjeffw (changes requested):
There's a couple of small changes in the review comments I'd like you to make.

Also try to remove the merge conflicts indicated in the PR.

src/templates/actor/tracker-manager.html
rinickolous: Added a commit which renames all *.html files to *.hbs and updates references accordingly.

test/parselink.test.js
rinickolous: When I added this I was thinking about some sort of module I saw a while back that let you simulate the FoundryVTT environment without a full instance of it running, so these global objects would work. However, unless that becomes something more concrete (i.e. I find the thing and decide it's good enough to propose using) , I agree that it's not really a hack. Comment removed in commit.

rinickolous:
I think I screwed up the merges and made this very messy so I'm likely going to resubmit this PR and just re-do the changes after the bugfixes and formatting commits have already been accounted for.

rinickolous:
I disabled the non-libWrapper use of patchFunc temporarily because of the way the `eval()` function interacts with rollup bundling.
I think ideally we just shouldn't wrap functions in this manner at all, and think that in the one instance in which patchFunc is currently used (displaying actor source info as text), we can re-create this functionality as part of the ActorSheet migration. Alternatively, I can try to re-create this functionality now to make sure the functionality is never removed (for non-libWrapper users).

@mjeffw please let me know what you think about this. I can undo this change if necessary and find another way around.

rinickolous:
Left to do: Include compatible release and pre-release GitHub workflows

mjeffw:
Thinking more about this PR and I think we need to modify or delay it a bit. Once we merge this PR into r18, it will be very hard to merge up changes from bugfix/0.17.20 due to the reorganization of the files. Should we keep the files where they are for now, and add generated files to .gitignore until there is no more Foundry V12 work to fix/enhance?

Remember we don't want perfection, we want "good enough" while keeping everything usable and used by our community.

test/parselink.test.js
mjeffw: I'll be pedantic here, so forgive me. My definition of unit tests are the tests that can run while we are editing files in the IDE, so that we get feedback instantly when we break something. It would also be beneficial to have integration or component tests, which are run in a production-like environment. The unit tests are almost more important, because the faster we get feedback, the lower the effort to keep the code clean and running.

rinickolous: Thanks for that. I want to avoid using terms incorrectly so I appreciate it.

rinickolous:
I'm ok with delaying in the short term but I also think the sooner we have a working type system to help navigate the API the easier it will be to find issues and add new features.

I could also make the same change on the V12 branch. It's very straightforward to do, though I also appreciate if you're reluctant to authorise that for what should amount to only a patch release.

In any case, the change is almost ready, whenever we're ready for it. Once I add the last component (GitHub workflows), we can just slot it into a release.

rinickolous:
Now that I think about it, a good approach which would keep files in-place would be to change the tsconfig and roll up settings to produce generated files in the dist folder while also keeping source files in-place. That would resolve the merge issues you mentioned.

I don't think it's as clean as moving files to src, but we can definitely just do that part at a later point.

I'll add a commit to do this when I get home shortly.

mjeffw:
I don't want to delay anything, either make the same change in 0.17.20 or don't rearrange the code. Let's get it merged sooner.

rinickolous:
Ok, files have been moved back (except for HTML files renamed with HBS extensions, though this can be reversed if you think it'd be an issue), and workflows have been updated to account for the build step.

The only thing remaining is fixing patchFunc. Rollup recommends not using eval because the way it bundles code moves all modules into the same scope, meaning any local variables are out of scope. [Here's the relevant page on their wiki](https://rollupjs.org/troubleshooting/#avoiding-eval).

I can try to manually amend the one function it is currently being used for so we don't lose the associated functionality, but using Rollup (or Vite, which I also just tested) doesn't seem to be at all compatible with the use of eval with locally scoped variables.

mjeffw:
Seems like we could just do what is recommended on the rollup page and copy eval->eval2. 

BYW, I'm not sold on minification of the code in any case.

mjeffw:
What is the impact of not using rollup at all?

rinickolous:
> What is the impact of not using rollup at all?

it (or another bundler) is necessary to build the code as far as I know.

rinickolous:
> Seems like we could just do what is recommended on the rollup page and copy eval->eval2.
> 
> BTW, I'm not sold on minification of the code in any case.

Using eval2 doesn't resolve the issue because patchFunc references a locally scoped variable.

Having just tested it now, we _could_ implement a globally scoped object, maybe something like a Record of unique identifiers and their corresponding functions. That should work.

rinickolous:
let me know what you think of this proposed solution. If it's fine then I think this thing is ready to go.

lib/change-log.js
```diff
@@ -28,7 +28,7 @@ export class ChangeLogWindow extends FormApplication {
     let data = await super.getData()
 
     let xhr = new XMLHttpRequest()
-    xhr.open('GET', 'systems/gurps/changelog.md')
+    xhr.open('GET', 'systems/gurps/assets/changelog.md')
```
mjeffw: Why this move to assets/?

rinickolous: Oh, that's residual from when I previously moved the change log to the src folder. I'll undo it.

rinickolous:
Ok, that's that sorted. I also pushed another commit which adds in the root SASS file. This required moving an image referenced in the CSS files to a different location (the hex-colormap image). Let me know if this is in the _wrong_ place.

utils/debugger.js
```diff
@@ -1,11 +1,14 @@
...
+
 // Copied from Monks Little Details module
 export let patchFunc = (prop, func, type = 'WRAPPER') => {
   let nonLibWrapper = () => {
-    const oldFunc = eval(prop)
+    const id = foundry.utils.randomID()
+    _patchedFuncs[id] = eval(prop)
     eval(`${prop} = function (event) {
-            return func.call(this, ${type != 'OVERRIDE' ? 'oldFunc.bind(this),' : ''} ...arguments);
+            return func.call(this, ${type != 'OVERRIDE' ? `_patchedFuncs["${id}"].bind(this),` : ''} ...arguments);
```
mjeffw, line 11: Explain what this does to enable eval functionality? Just curious what is happening here.

rinickolous: This does not enable eval functionality _in general_, but preserves its current uses.

GGA uses eval in 3 places:
1. During GCA actor import when resolving DR values expressed as arithmetic functions. This is unchanged and needs no immediate amending.
2. In patchFunc, when resolving the name of a function to be patched. This also needs no amending.
3. In patchFunc, when setting the aforementioned function to be patched to a wrapped version of itself which also executes additional code.

Since basically all modern bundlers put all modules in the same scope, it causes `eval` to lose access to the locally declared variable `oldFunc`, since it is being executed in the global scope instead. To fix this, `eval` would have to reference a variable which exists in the global scope. To that end, I declared a global scope `_patchedFuncs: Record<string,Function>`, and am using Foundry's default id generation function to assign random ids to patched functions. I think random IDs are appropriate in this context because they avoid potential issues caused by arrays (where the index a function sits at may for some reason be different than expected). The need for a Record rather than single variable assigned to a function should also be obvious: If we use patchFunc multiple times, the variable would be overridden as many times as it is run.

A potential alternative to declaring the _patchedFuncs variable in the global scope would be to have _patchedFuncs be a property of a globally accessible singleton class.

mjeffw (commented):
What if we don't jump all the way to using a bundler like rollup yet, and just use tsc and cpx to build the code and create the dist folder?

The advantages may be that 
- We don't need to touch patchFunc yet, since we have not bundled the code into the same namespace
- The output is exactly the same as the current pure javascript approach.
- The differences between a 0.17.x release and 0.18.x would be minimal, allowing for less cognitive load when moving from one to the other.

I did an experiment with this and here's what I ended up with:

package.json:
```
{
  "name": "gurps",
  "version": "0.9.0",
  "description": "NaN's GURPS Game Aid",
  "type": "module",
  "scripts": {
    "test": "jest --coverage",
    "tdd": "jest --coverage",
    "handlebars": "handlebars",
    "prettier": "prettier --write .",
    "start": "node dist/index.js",
    "build": "tsc && npm run copy-assets",
    "copy-assets": "cpx 'templates/**/*' dist/templates && cpx 'styles/**/*' dist/styles && cpx 'assets/**/*' dist/assets && cpx 'icons/**/*' dist/icons && cpx 'exportutils/**/*' dist/exportutils && cpx 'ui/**/*' dist/ui && cpx 'changelog.md' dist/ && cpx 'LICENSE.txt' dist/ && cpx 'README.md' dist/ && cpx 'system.json' dist/ && cpx 'template.json' dist/ && cpx 'utils/*.*/*' dist/utils"
  },
  "devDependencies": {
    "@babel/core": "^7.26.0",
    "@babel/preset-env": "^7.26.0",
    "@league-of-foundry-developers/foundry-vtt-types": "github:League-of-Foundry-Developers/foundry-vtt-types#documents-v2",
    "@types/node": "^22.14.1",
    "babel-jest": "^29.7.0",
    "cpx": "^1.5.0",
    "handlebars": "^4.7.7",
    "jest": "^29.7.0",
    "prettier": "^3.5.3",
    "typescript": "^5.8.3"
  }
}
```

tsconfig.json:
```
{
  "compilerOptions": {
    "target": "ES2020" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
    "module": "ESNext" /* Specify what module code is generated. */,
    "rootDir": "./" /* Specify the root folder within your source files. */,
    "moduleResolution": "Node" /* Specify how TypeScript looks up a file from a given module specifier. */,
    "resolveJsonModule": true /* Enable importing .json files. */,
    "allowJs": true /* Allow JavaScript files to be a part of your program. Use the 'checkJS' option to get errors from these files. */,
    "checkJs": false /* Enable error reporting in type-checked JavaScript files. */,
    "outDir": "./dist" /* Specify an output folder for all emitted files. */,
    "esModuleInterop": true /* Emit additional JavaScript to ease support for importing CommonJS modules. This enables 'allowSyntheticDefaultImports' for type compatibility. */,
    "forceConsistentCasingInFileNames": true /* Ensure that casing is correct in imports. */,
    "strict": true /* Enable all strict type-checking options. */,
    "skipLibCheck": true /* Skip type checking all .d.ts files. */
  },
  "include": ["./**/*"],
  "exclude": [
    "node_modules",
    "dist",
    "src",
    "dev-utilities",
    "test",
    "jest.config.js",
    "tsconfig.json",
    "package-lock.json",
    "package.json",
    "plans.md",
    "prettier.config.cjs",
    "ReleaseProcess.md"
  ]
}
```

After the 0.17.x release is complete, we can upgrade to using rollup at that time.

rinickolous:
Looks great. I'll just amend it to add a --watch option and push (after dinner)

rinickolous:
Ok, had a slight hiccup with tinycolorpicker but resolved that too.
Current npm functions are as follows:

- clean - cleans dist folder
- build:code - one time code build
- build:styles - one time style build
- watch:code - runs code build with watcher
- watch:styles - runs styles build with watcher
- copy-assets - same as yours above
- watch - runs watch:code, watch:styles, copy-assets in parallel
- build - runs clean, build:code, build:styles, copy-assets sequentially
- dev - runs clean, then watch sequentially

We'll likely just be using build and dev in practice for now

---

I think it's ready to go.

mjeffw (approved):

Issue #2204, opened 2025-04-22
Quadtree check for damage drop works not correctly with tokens with shiftet hitbox from GURPS Token Shapes and Movment


StefanLeng:
To get GURPS-Like Token movement with bigger tokens in the GURPS Token shape an movement module, i have to shift the center oft the Token for token > 1 hex, so that the center is in the front hex of the border and the hit area. Unfortunally, that moves part of the hit area out of the token bounds as recognized by the Quattree. So part of the hit area are not recocnized. 

see: https://github.com/StefanLeng/gurps-gridless/issues/5 

Teh schifted hit area works without problems for foundry functions like targeting or selecting, so i assume foundry dosn't uses the quuattree.
I think we have to do something  like 

let selectetTokens = canvas.tokens.objects.children.filter(t => t.hitArea.contains(dropData.x - t.x, dropData.y - t.y))

I think it is OK, because  the number of tokens per scene will not be big and the check is not that expensive.

PR #2205, opened 2025-04-22
Make damage drop compatible with GURPS Token Shapes and Movement

StefanLeng:
fixes https://github.com/crnormand/gurps/issues/2204

### DIFF
--- a/module/damage/damagechat.js
+++ b/module/damage/damagechat.js
@@ -447,16 +447,8 @@ export default class DamageChat {
       return false
     }
 
-    // Create a "reactangle" to represent the drop coordinates. It is really a point as width and height are 0.
-    const rect = new PIXI.Rectangle(dropData.x, dropData.y, 0, 0)
-
-    // Get all tokens under the point.
-    let selectedTokens = canvas.tokens.quadtree
-      .getObjects(rect, {
-        collisionTest: o => o.t.hitArea.contains(dropData.x - o.t.x, dropData.y - o.t.y),
-      })
-      .values()
-      .toArray()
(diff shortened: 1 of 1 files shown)


module/damage/damagechat.js
```diff
@@ -447,16 +447,8 @@ export default class DamageChat {
...
-    const rect = new PIXI.Rectangle(dropData.x, dropData.y, 0, 0)
-
-    // Get all tokens under the point.
-    let selectedTokens = canvas.tokens.quadtree
-      .getObjects(rect, {
-        collisionTest: o => o.t.hitArea.contains(dropData.x - o.t.x, dropData.y - o.t.y),
-      })
-      .values()
-      .toArray()
+    // Get all tokens under the drop point.
```
mjeffw, line 450: It does bother me that your module breaks the quad tree algorithm, as that is how the canvas efficiently draws its contents, determines what is under a point, and other functions. There is a potential for your module to cause other, subtler bugs. I'll allow it but I'm keeping the original quad tree code commented out for now.

mjeffw (approved):

module/damage/damagechat.js
StefanLeng: Yes, that is bothersome. I just have not found anohter way to make a token snap to the grid with its head instead of its center, I am generally not happy how deep I had to change the way tokens are  drawn to make he GURPS Movement work on foundry hex grids (for larger tokens).
On the other hand, I have extensivly tested all foundry functionality that target tokens and everything seems to work. But the potiential for subtle bugs remain,

PR #2207, opened 2025-04-22
Auto-formatted all code with prettier, added prettier as dependency

rinickolous:
To avoid the issue which has come up as part of my previous two PRs, I ran prettier on the repo (and added it as a dev dependency) to auto-format all repo code.

Note: Some handlebars code has not been formatted as handlebars support for prettier seems to be very limited. It currently uses the Glimmer formatter which doesn't support some Handlebars nodes.

Ideally, we would just use a workflow or something to do this automatically.


mjeffw (approved):

mjeffw:
Please make the same change in the bugfix/0.17.20 branch so we minimize problems when merging changes up to the r18 branch.

mjeffw:
Please make the same change in the bugfix/0.17.20 branch so we minimize problems when merging changes up to the r18 branch.

rinickolous:
> Please make the same change in the bugfix/0.17.20 branch so we minimize problems when merging changes up to the r18 branch.

I'm not sure I understand. This change is for 0.17.20. Did you mean to say that I should make the same changes in the r18 branch?

mjeffw:
Sorry, I'm reviewing this from my phone and I missed the branch you were merging into. I'm visiting my son in Washington state.

PR #2230, opened 2025-04-27
Build Tool Improvements

rinickolous:
I changed the way our build tools work somewhat to speed up development especially when working with incremental changes to static files.

The change cleands up the static file copy process in the package.json file by moving the functionality to a dedicated `build` folder and adds a watcher daemon for static files, allowing incremental and frequent changes to static files without having to recompile anew.

This change also includes a change to `fvtt-types`. The `documents-v2` branch was merged into `main` this morning, so the most up-to-date branch is now `main`.


mjeffw:
> This change also includes a change to `fvtt-types`. The `documents-v2` branch was merged into `main` this morning, so the most up-to-date branch is now `main`.

I don't see any merges into `main` other the last one I did on April 22. In any case, don't jump ahead on any migrations/upgrades without discussion.

package.json
mjeffw, line 1: Rather than "roll-your-own" solution, did you look into the `cpx` documentation? It integrates with `watchify` which may have solved the problem without custom code.

rinickolous: Thanks for catching me on this. No, I didn't previously look at the cpx documentation, and it looks like cpx offers a far less complicated solution than the one I used. I reversed the dependency changes and added cpx back in. The current solution still has one custom file but all it does is make it easier to see which directories and files are being copied at a glance.

tsconfig.json
```diff
@@ -19,7 +19,7 @@
 
     /* JavaScript Support */
     "allowJs": true /* Allow JavaScript files to be a part of your program. Use the 'checkJS' option to get errors from these files. */,
-    "checkJs": false /* Enable error reporting in type-checked JavaScript files. */,
+    "checkJs": true /* Enable error reporting in type-checked JavaScript files. */,
```
mjeffw, line 22: Should this be false until we can rewrite the JS? Or can we flag the problems but show them as warnings instead of errors? I don't like all the errors in my IDE and I felt like it stopped a multi-step build from progressing past the transpilation step.

rinickolous: Makes sense to me. I think most of the errors just amount to complains about lack of non-null and non-undefined checks and implicitly any types, all of which can be dealt with during the conversation process.

PR #2231, opened 2025-04-27
Migrate GurpsActorSheet to AppV2

rinickolous:
This change adds two new versions of GurpsActorSheet.

The first is an experimental from the ground up rewrite of GurpsActorSheet, aiming to match Foundry's AppV2 visual style and clean up functionality. This sheet is currently unfinished and disabled for users.

The second is a minimal migration of GurpsActorSheet to AppV2, aiming to imitate the AppV1 functionality as closely as possible while migrating to the AppV2. This is also currently unfinished. I will mark this branch as ready for merge when this migration is complete.


mjeffw:
I'd like the complete rewrite removed from this PR (perhaps move all of that code to another branch). I don't want any code in the r18 branch that isn't going to be used as soon as the branch is published. 

What I'd like you to do is:

- Migrate the existing `actor-sheet.js` code to ApplicationV2. Yes, this means do not migrate to TS.
- You might make structural changes required for ApplicationV2 without updating the event handling code at all. Keep it in jQuery using `const html = $(this.element)` for backward compatibility. Make sure the code still works. Submit that as a PR.
- If possible, do a piecemeal migration of the code in `activateListeners` to use native HTML API. That means picking one or two sections of that code that is currently using jQuery, testing the current actor sheet for that specific functionality, moving it to use the HTML API, and retesting it. Once you get a one or a few sections of code updated (not the whole thing!) submit each one of these tiny steps as a PR.

rinickolous:
I understand the reasoning behind all of these requests except the "no TS" one. If the build tools specifically intended for the purpose of moving things over to TS are available, why should I not use them?

mjeffw:
The goal is to change as little code as possible while we are keeping two versions active.

rinickolous:
That's fine in theory but it's simply not going to happen when migrating large classes to ApplicationV2. There  are large structural differences between AppV1 and AppV2, we're not going to get a clean merge for something as big as an actor sheet either way. It just seems like I'm being asked to stunt my developer experience arbitrarily.

mjeffw:
Maybe I should do an App -> AppV2 first, and then my expectations will be set appropriately. Also, maybe the actor sheet is too big to start with.

Okay, hold off on doing any more until I at least think about it, if not actually migrate a smaller app.

rinickolous:
I'm perfectly capable of doing this Actor sheet migration myself. I'm trying to help with the workload, but you need to trust me enough to actually let me work.

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

mjeffw:
I trust you as much as anyone. In my experience, one should avoid wholesale migrations which require rewriting large amounts of code as an indivisible unit (unless one has full test automation coverage). Instead, one should do the migration a piece at a time and at each step prove that the thing still works. This reduces the amount of testing required at each step and reduces the number of hidden bugs that come out of such a migration. 

It could even be beneficial to create a new top-level AppV2 class that calls into the existing App class methods to minimize the code changes. 

I need to get some experience doing this migration so I can suggest the best approach.

rinickolous:
With all due respect, the suggestions in this last comment seem like unnecessary additional work. I'm experienced enough with AppV2 to be able to do the minimal migration cleanly and keep the change from getting out of scope for the sale of compatibility.

I have somewhere to be today but I'm more than willing to go with your original plan (minus the no Typescript request), split up the work as requested and submit the different aspects for review individually while also doing sufficient testing myself. I'm not sure how I could unit test work like this (or even if that's appropriate) considering it's all UI interactions but if you can point me in the right direction that would be great.

mjeffw:
Okay, let's give that a try. I don't have a way to do test automation (not unit, but integrated or functional testing) yet. It will be a lot of work (doing something like automating through the browser using selenium or something similar). I was just pointing out that without that test automation, making more than a handful of changes at a time will be difficult to test.

rinickolous:
Sure, that makes sense. Thanks for hearing me out on this, I appreciate it.

PR #2233, opened 2025-04-27
2228 v13 deprecation dialog   dialogv2 for actorjs

mjeffw:
Rather than migrate what was essentially the same code twice, I refactored this method to have only one call to the new DialogV2.prompt.


rinickolous (approved):
Looks good, no issues that I can see.

PR #2235, opened 2025-04-29
convert to DialogV2 for _splitEquipment

mjeffw:
I ended up delegating to the same DialogV2 as the previous two PRs.


rinickolous (commented):
No issues, merging.

PR #2238, opened 2025-05-02
Adjust Modifier Bucket position based on user settings

mjeffw:


rinickolous (changes requested):
The "right" option give the bucket button too wide a gap on the left, causing the dice button to be partially cut off when the sidebar is expanded. This only affects uses running Foundry at or close to the minimum supported resolution of 1366x768 px, but we should still account for this.
image

I would suggest overriding the margin (left and right) values for when the #bucket-container element is pinned to the right.

mjeffw:
I could never prevent it from overlapping. This is at least partially if not entirely because of how Foundry positions the hotbar and the chatbar. There is a workaround -- if the modifier bucket and chatbar input overlap, expand the right side area and it will fix itself.

PR #2242, opened 2025-05-04
Implement migration system for damage chat messages and update chat m…

### LINKED ISSUE #2240: ChatMessage flags are being created incorrectly

mjeffw:
…essages to use new flag. Fixes #2240.


rinickolous (approved):
My only real note here is that for future migrations of Documents, we may want to use the `Document#_stats` object to check if the Document is a candidate for migration, which can lead to early returns without needing to check against a series of values in cases where said values can be cumbersome to track. This doesn't really apply in this case since it's such a simple migration, but something to consider for any future migrations.

Approved.

PR #2243, opened 2025-05-04
Fixed ruler incompatibility with Foundry v13

### LINKED ISSUE #2241: v13: The canvas ruler doesn't work as before
The canvas ruler no longer adds to the modifier bucket.
There was also a change in Foundry v13 which causes the ruler to no longer highlight hexes when measuring, but this can probably be ignored.

rinickolous:
I reimplemented the Ruler class, seeing as the functions which we overrode for distance modifiers have been changed in Foundry v13. Support for DragRuler has been removed, but DragRuler has also been made redundant with Foundry v13 as Foundry now shows the distance travelled by token by default.

This pull request also includes a relatively comprehensive utility file for dealing with length measurements. This was lifted/adapted from GCS and was basically just copied over from the [gcs-datamodel](/crnormand/gurps/tree/feature/gcs-datamodel) branch which I have been working on separately. This utility is used as a replacement for the function which was previously used to calculate distance in yards, since I already wrote it.

This resolves #2241

### DIFF
--- a/module/canvas/interaction/index.ts
+++ b/module/canvas/interaction/index.ts
@@ -1,1 +1,0 @@
-export * from './ruler.js'
--- a/module/canvas/interaction/index.ts
+++ b/module/canvas/interaction/index.ts
@@ -1,0 +1,1 @@
+export * from './ruler.js'
--- a/module/data/common/length.ts
+++ b/module/data/common/length.ts
@@ -76,6 +76,8 @@ class Length<Parent extends DataModel.Any | null = DataModel.Any | null> extends
     [Length.Unit.Feet]: 12,
     [Length.Unit.Yard]: 36,
     [Length.Unit.Mile]: Length.INCHES_PER_MILE,
+    // In GURPS simplified lengths, 1 millimeter is defined as 0.36 inches (36/100),
+    // which is a non-standard abstraction used for gameplay purposes.
     [Length.Unit.Millimeter]: 36 / 100,
     [Length.Unit.Centimeter]: 36 / 1000,
     [Length.Unit.Kilometer]: 36000,
--- a/module/data/common/length.ts
+++ b/module/data/common/length.ts
@@ -76,10 +76,8 @@ class Length<Parent extends DataModel.Any | null = DataModel.Any | null> extends
     [Length.Unit.Feet]: 12,
     [Length.Unit.Yard]: 36,
     [Length.Unit.Mile]: Length.INCHES_PER_MILE,
-    // In GURPS simplified lengths, 1 millimeter is defined as 0.36 inches (36/100),
-    // which is a non-standard abstraction used for gameplay purposes.
-    [Length.Unit.Millimeter]: 36 / 100,
-    [Length.Unit.Centimeter]: 36 / 1000,
+    [Length.Unit.Millimeter]: 36 / 1000,
+    [Length.Unit.Centimeter]: 36 / 100,
     [Length.Unit.Kilometer]: 36000,
     [Length.Unit.Meter]: 36,
     // Simplified to 93 million miles
--- a/module/data/common/length.ts
+++ b/module/data/common/length.ts
@@ -1,0 +1,218 @@
+import DataModel = foundry.abstract.DataModel
+import fields = foundry.data.fields
+
+enum LengthUnit {
+  FeetAndInches = 'ft_in',
+  Inch = 'in',
+  Feet = 'ft',
+  Yard = 'yd',
+  Mile = 'mi',
+  Millimeter = 'mm',
+  Centimeter = 'cm',
+  Kilometer = 'km',
+  Meter = 'm',
+  AstronomicalUnit = 'au',
+  Lightyear = 'ly',
+  Parsec = 'pc',
+}
+
+/* ---------------------------------------- */
+
+const lengthSchema = () => {
+  return {
+    value: new fields.NumberField({ required: true, nullable: false, min: 0, initial: 0 }),
+    unit: new fields.StringField({
+      required: true,
+      nullable: false,
+      choices: Object.values(LengthUnit),
+      initial: LengthUnit.Inch,
+    }),
+  }
+}
+
+type LengthSchema = ReturnType<typeof lengthSchema>
+
+/* ---------------------------------------- */
+
+class Length<Parent extends DataModel.Any | null = DataModel.Any | null> extends DataModel<LengthSchema, Parent> {
+  /* ---------------------------------------- */
+  /*  Static Properties                       */
+  /* ---------------------------------------- */
+
+  static Unit = LengthUnit
+
+  /* ---------------------------------------- */
+
+  static INCHES_PER_MILE = 63360
+  /* ---------------------------------------- */
+
+  // Maximum number of decimal places to display
+  static ROUNDING_PRECISION = 4
+
+  /* ---------------------------------------- */
+
+  static UNIT_LABELS: Record<LengthUnit, string[]> = {
+    // Feet & Inches is treated as a special case which uses the x'y" format
+    [Length.Unit.FeetAndInches]: [],
+    [Length.Unit.Inch]: ['in', 'inch', 'inches'],
+    [Length.Unit.Feet]: ['ft', 'foot', 'feet'],
+    [Length.Unit.Yard]: ['yd', 'yard', 'yards'],
+    [Length.Unit.Mile]: ['mi', 'mile', 'miles'],
+    [Length.Unit.Millimeter]: ['mm', 'millimeter', 'millimeters'],
+    [Length.Unit.Centimeter]: ['cm', 'centimeter', 'centimeters'],
+    [Length.Unit.Kilometer]: ['km', 'kilometer', 'kilometers'],
+    [Length.Unit.Meter]: ['m', 'meter', 'meters'],
+    [Length.Unit.AstronomicalUnit]: ['au', 'astronomical unit', 'astronomical units'],
+    [Length.Unit.Lightyear]: ['ly', 'lightyear', 'lightyears'],
+    [Length.Unit.Parsec]: ['pc', 'parsec', 'parsecs'],
+  }
+
+  /* ---------------------------------------- */
+
+  // Everything is relative to the inch, using GURPS simplified lengths only
+  static UNIT_CONVERSIONS: Record<LengthUnit, number> = {
+    [Length.Unit.FeetAndInches]: 1,
+    [Length.Unit.Inch]: 1,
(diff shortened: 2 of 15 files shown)


mjeffw:
Sounds good!

BTW, I don't want you implementing the GCS data model by yourself because I need to understand it. Plus there may be some need of conversation about the approach we should take. If you completely do it it will be a mystery to me.

rinickolous:
Of course. Now that you say it, I've actually already made a deviation when it comes to how lengths are stored compared to GCS, because I think the way it was doing it is maybe not the most appropriate for our use case.

Also, I haven't started on the datamodel proper, so far I've just implemented utility functions for length, weight, and am about halfway through implementing the functions GCS uses for evaluating attribute values. No structural stuff yet.

Just to confirm, this is fine to merge?

mjeffw:
Yah, let me review it and I'll merge. On vacation this week but I have my laptop.

Farmeroz:
A quick comment here - a lot of us are using Elevation Ruler (mostly after Drag Ruler wasn't updated for V12 for a very long time).  Have you checked compatibility there or confirmed redundancy in favour of V13 native ruler?  Speed colouring, pathfinding, and elevation measurement?

rinickolous:
> A quick comment here - a lot of us are using Elevation Ruler (mostly after Drag Ruler wasn't updated for V12 for a very long time).  Have you checked compatibility there or confirmed redundancy in favour of V13 native ruler?  Speed colouring, pathfinding, and elevation measurement?

Haven't checked yet, but thanks for the heads up. I'll check in a bit.

Farmeroz:
> > A quick comment here - a lot of us are using Elevation Ruler (mostly after Drag Ruler wasn't updated for V12 for a very long time).  Have you checked compatibility there or confirmed redundancy in favour of V13 native ruler?  Speed colouring, pathfinding, and elevation measurement?
> 
> Haven't checked yet, but thanks for the heads up. I'll check in a bit.

Thanks!

module/canvas/interaction/index.ts
mjeffw, line 1: I'm not sure there's any meaning to having an `interaction` folder in the canvas module. I think most code in canvas will be to enable some form of interaction. Could or should this be flattened out into simply the `canvas` folder?

Oh, I see, we are mirroring the foundry's own module structure (`foundry.canvas.interaction.Ruler`). I might be okay with that.

module/data/common/length.ts
```diff
@@ -0,0 +1,220 @@
...
+      initial: LengthUnit.Inch,
+    }),
+  }
+}
+
+type LengthSchema = ReturnType<typeof lengthSchema>
+
+/* ---------------------------------------- */
+
+class Length<Parent extends DataModel.Any | null = DataModel.Any | null> extends DataModel<LengthSchema, Parent> {
```
mjeffw, line 37: Why make this extend DataModel? Are there uses of Length that are independent of the DataModel?

All of the conversion stuff seems independent of data persistence, for example. I'd consider breaking out the the behavior of Length and Unit into a non-Foundry dependent class, and then make it persistent by defining a DataModel length that uses the independent Length and Unit stuff.

module/data/common/length.ts
```diff
@@ -0,0 +1,220 @@
...
+  static UNIT_CONVERSIONS: Record<LengthUnit, number> = {
+    [Length.Unit.FeetAndInches]: 1,
+    [Length.Unit.Inch]: 1,
+    [Length.Unit.Feet]: 12,
+    [Length.Unit.Yard]: 36,
+    [Length.Unit.Mile]: Length.INCHES_PER_MILE,
+    // In GURPS simplified lengths, 1 millimeter is defined as 0.36 inches (36/100),
+    // which is a non-standard abstraction used for gameplay purposes.
+    [Length.Unit.Millimeter]: 36 / 100,
+    [Length.Unit.Centimeter]: 36 / 1000,
```
mjeffw: Doesn't this make a centimeter (36/1000 = 0.036 inches) smaller than  a millimeter (36/100 = 0.36 inches)? Are these two units swapped?

module/data/common/length.ts
```diff
@@ -0,0 +1,220 @@
...
+    [Length.Unit.Kilometer]: ['km', 'kilometer', 'kilometers'],
+    [Length.Unit.Meter]: ['m', 'meter', 'meters'],
+    [Length.Unit.AstronomicalUnit]: ['au', 'astronomical unit', 'astronomical units'],
+    [Length.Unit.Lightyear]: ['ly', 'lightyear', 'lightyears'],
+    [Length.Unit.Parsec]: ['pc', 'parsec', 'parsecs'],
+  }
+
+  /* ---------------------------------------- */
+
+  // Everything is relative to the inch, using GURPS simplified lengths only
```
mjeffw, line 72: Is the plan to store everything in inches and convert on display?

module/gurps.js
```diff
@@ -50,7 +50,8 @@ import { colorGurpsActorSheet } from './color-character-sheet/color-character-sh
 
 import HitFatPoints from '../lib/hitpoints.js'
 import Initiative from '../lib/initiative.js'
-import { GURPSRange, RulerGURPS, setupRanges } from '../lib/ranges.js'
+import { GURPSRange, setupRanges } from '../lib/ranges.js'
```
mjeffw, line 53: The model I'd like to move towards is most/all of the code in modules, and in `gurps.js` (or `gurps.ts` in the future) it calls an initialization method in the module, which initializes/constructs new module specific stuff and updates the global `gurps` object, rather than all this happening in `gurps.js`.

rinickolous:
> > > A quick comment here - a lot of us are using Elevation Ruler (mostly after Drag Ruler wasn't updated for V12 for a very long time).  Have you checked compatibility there or confirmed redundancy in favour of V13 native ruler?  Speed colouring, pathfinding, and elevation measurement?
> > 
> > Haven't checked yet, but thanks for the heads up. I'll check in a bit.
> 
> Thanks!

Quick update on this: Elevation ruler has not yet been updated for v13, so I can't yet do anything about making sure it's compatible

module/gurps.js
rinickolous: makes sense. just so we're on the same page, how would you have me do that in this instance?

module/data/common/length.ts
rinickolous: Good catch, I'll get that fixed.

module/data/common/length.ts
rinickolous: I'm sort of undecided on this. The way GCS stores lengths appears to be:
- Lengths in the JSON files are stored as strings with a user-inputted unit.
- Lengths in memory are stored as 4dp fixed point numbers, in inches.

But GCS also has its own reasons for doing so which I don't think carry over to GGA. I believe storing lengths identically to how GCS does it in the persistent data is fine, while instantiated lenghts would likely be better suited as objects with their own instance properties if we need to do things like display or calculate anything.

module/canvas/interaction/index.ts
rinickolous: That's one of my primary reasons for the file organisation structure I'm proposing. I think it would be beneficial to have GGA's namespace structure mirror Foundry's own so it's more intuitive for anyone making modules for GGA. And if the namespace should mimic Foundry's own, I reckon it follows pretty naturally that the file strucutre should reflect the namespace organisation.

module/data/common/length.ts
rinickolous: I'm not sure if there are any uses of Length which will actually be independent of DataModel (I only see it being used for actor lengths/heights as well as ruler measurement for SSRT reasons etc.). Extending DataModel mostly serves the purpose of getting to use Foundry's own data validation methods when doing anything with Length. Even so, we *can* move the utility functionality of Length to some sort of Length Utility space separate from anything Foundry-dependent, then have the Length DataModel class simply call these utility functions from its own instance functionality, but I don't see why that would be necessary at this stage.

module/data/common/length.ts
rinickolous: To be clear, the way things are currently stored is in objects with structure `{value: number, unit: Length.Unit}`. I think this helps to step around issues which would arise by storing things in floats representing value in inches.

GCS deals with the float issue by using its own fixed point number library, but I think this is likely to be cumbersome and unnecessary for our use case.

module/gurps.js
mjeffw: The way I was experimenting with was to implement an  `init` function in the module's `index.js` file, and have that method use hooks to add stuff to the GURPS global object, or to hook into other events. See `module/damage/index.js` as an example.

BTW, this is a place where organizing the GURPS object into more of a hierarchy would make sense: Perhaps all the damage functions would go under `GURPS.damage.*`. It will be disruptive to our macro-writing users, though.

rinickolous: Sounds good, I'll adjust to match the same format.

Regarding the macro-writing users: I think moving to a hierarchical GURPS object would be good long-term, but I think the best way of going about it would be to:
1. Figure out some sort of structure from the get-go.
2. Adjust things to reflect that structure, wrapping any flat-object keys into a wrapper function with a generous deprecation warning attached.
3. Remove the flat keys at a later point.

rinickolous:
Ok, made some adjustments. I think some of the questions/discussions in here are still unresovled. To summarise:
- Length is currently a class extending DataModel (for its built-in validation functions and so that it can be easily used in fields when we start moving our Document system data towards extending DataModel in the future). We *can* change this but it's just as easy to do so later as now, and I don't see a need to do so immediately.
- The Length data is currently an object of type `{value: number, unit: Length.Unit}` where `Length.Unit` is an enum representing various conventional units of length.
- Lengths are calculated using GURPS simplified metric conversions, with all conversions expressed in inches. The values themselves are not stored in just inches at any point, but we can change how this works if there's some real need for it.
- The Ruler now lives in a module `Canvas` as requested.

Let me know if there's anything left actually unresolved, @mjeffw .

Farmeroz:
> > > > A quick comment here - a lot of us are using Elevation Ruler (mostly after Drag Ruler wasn't updated for V12 for a very long time).  Have you checked compatibility there or confirmed redundancy in favour of V13 native ruler?  Speed colouring, pathfinding, and elevation measurement?
> > > 
> > > 
> > > Haven't checked yet, but thanks for the heads up. I'll check in a bit.
> > 
> > 
> > Thanks!
> 
> Quick update on this: Elevation ruler has not yet been updated for v13, so I can't yet do anything about making sure it's compatible

Noticed this issue raised on the related module's GitHub:

https://github.com/caewok/fvtt-elevation-ruler/issues/257

The main thing that's nice about Elevation Ruler is pathfinding, but it's not critical if the internal implementation allows the user to map a path themselves (rather than just a straight line), and it has the expected colours and respects current move and so on.

PR #2244, opened 2025-05-06
Migrated all Actor sheets to ApplicationV2

rinickolous:
This is sort of a second attempt at migrating GurpsActorSheet. It's more minimal than my last attempt in terms of any refactoring, i.e. almost all listeners remain as JQuery listeners for now. I haven't done *extensive* testing on this, but most things seem to be working with no issues.

I did notice that the context menu doesn't seem to be working for non-equipment items but this issue also persists for AppV1 documents. I can investigate this and add it to this PR if needed.

The original actor sheet functions remain in place, they're just not registered as I managed to replace all of them with AppV2 versions. Currently, GURPS.ActorSheets.character is *also* set to the new ActorSheetGURPS constructor. This may break the categorised sheet, but I already have a fork of the Gaming Ballistic sheet ready to ensure we don't fall out of compatibility with our modules.

As a final note, dark mode is sort-of supported in this branch for most sheets. I didn't spend much time on it and it has a few quirks but I want to run this thing by you before spending time fixing it.


module/applications/actor/actor-combat-sheet.ts
mjeffw, line 1: I don't like this code organization. Code should be organized to reflect the domain -- in this case a TTRPG -- and not the role of the components in the software. I would prefer (strongly) that we have an `applications` folder inside the `actor` module, and not have all the applications in one place regardless of what they represent. Likewise for other things like

mjeffw (changes requested):

module/applications/actor/actor-combat-sheet.ts
rinickolous: Your comment seems to be cut off. I can make the organisation changes if you want, just want to make sure I'm getting the full comment here.

PR #2248, opened 2025-05-13
Fixed aim bonus not resetting after being used once in combat.

### LINKED ISSUE #2153: ACC bonus for Aim only works once per combat
Using Tagged modifiers with auto-apply and roll confirmation, the first Aim-attack sequence in a given combat adds weapon ACC, but the second and any subsequent Aim-attack sequences, no matter how far apart they happen in the same combat, only adds +1 as if it were an extra turn of Aim.

Ending combat and then starting it again allows ACC Aim bonus to work again, but only on the first Aim-attack sequence.

May be related to #2152 

First Aim-attack sequences in combat:

Image

All subsequent Aim-attack sequences

Image

rinickolous:
The issue was caused by explicitly assigning `undefined` as a value rather than using `null`. There are a few other instances of this happening in the GGA code with as yet unknown consequences, so I would strongly recommend not doing so in the future.

This resolves #2153 .

### DIFF
--- a/module/token-actions.js
+++ b/module/token-actions.js
@@ -638,7 +638,6 @@ export class TokenActions {
         console.log(`Add +1 Move bonus to ${this.token.name})`)
         break
       case 'aim':
-        console.log(this.actor.name, 'Current Aim before:', { ...this.currentAim })
         Object.keys(this.currentAim).map(k => {
           const a = this.currentAim[k]
           if (a.startAt === null) a.startAt = this.currentTurn - 1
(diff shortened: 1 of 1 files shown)


module/token-actions.js
```diff
@@ -90,9 +90,9 @@ export class TokenActions {
     this.extraActions = 0
     this.rapidStrikeBonus = 0
 
-    this.canDefend = undefined
-    this.canAttack = undefined
-    this.canMove = undefined
+    this.canDefend = null
+    this.canAttack = null
+    this.canMove = null
```
mjeffw, line 95: Is there a preference to using `null` as opposed to `undefined`?

PR #2260, opened 2025-05-17
Cleaned up Token class, fixed maneuver-related issues.

rinickolous:
I moved the Token class to Typescript and simplified various functions to do with choosing and rendering maneuvers. The initial scope of this branch was just fixing #2250 but while doing so I found various instances of maneuver-setting functions being duplicated, using deprecated functionality, and being overly complicated.

Walkthrough of changes per file:
- global.d.ts - Updated types to reflect updated class.
- lib/miscellaneous-settings.js - Updated imports
- module/actor/actor-sheet.js - Removed reference to non-existent function `GurpsActor#toggleEffectByName`
- module/actor/actor.js
  - Removed redundant reference to `settings` where `Settings` is already imported from the same file.
  - Removed unused function `_renderAllApps`
  - Removed reference to non-existent function `GurpsActor#toggleEffectByName`
- Other files
  - Removed references to non-existent function `GurpsToken#setEffectActive`

I removed references to Drag Ruler because it hasn't been updated in years.
I also moved TokenGURPS into its own module alongside the corresponding HUD function.


global.d.ts
```diff
@@ -15,7 +15,7 @@ declare global {
   }
 
   interface PlaceableObjectClassConfig {
-    Token: typeof GurpsToken
+    Token: typeof TokenGURPS
```
mjeffw: We need to unify our naming schemes. On one hand we have "GurpsActor" and on the other "TokenGURPS". Let's try to obey the "law of least surprise" and name things the same way.

It's generally accepted that classes use UpperCamelCase -- on that we (almost!) agree. We disagree on GURPS vs. Gurps. I prefer "Gurps" of course, and the following website agrees with me:

[TS Syle Guide:Syntax:Identifiers](https://ts.dev/style/#identifiers).

(See the "Abbreviations" bullet point below the table.)

Whether GURPS/Gurps comes first or last feels pretty subjective to me. But the project has already adopted the convention of putting the base class name last and prepending "Gurps" or other qualifiers to it.

For these types of issues, I'd consider consistency to be trump preference. The TS Style guide [agrees](https://ts.dev/style/#consistency).

If we are worried about a name collision (GurpsToken previous to this release, and GurpsToken as the new TS implementation) we can find another way to resolve that (GurpsTokenV2 as an example).

module/token/object.ts
mjeffw, line 1: I'm not a fan of using the generic `object.ts` name of a file. If we follow this naming convention throughout the code, we'd have a few dozen or more files with exactly the same names. I would prefer naming the file to be more specific, such as `gurps-token.ts`. 

The only exception I am currently making to this guidance is `index.ts` or `index.js` to denote the entry point of the module. Those files should have (almost) no code not related to initialization or exporting public entities from the module.

lib/miscellaneous-settings.js
```diff
@@ -3,7 +3,7 @@
 import { ItemImageSettings } from '../module/actor/actor-sheet.js'
 import { TaggedModifierSettings } from '../module/actor/effect-modifier-popout.js'
 import ModifierBucketJournals from '../module/modifier-bucket/select-journals.js'
-import { QuickRollSettings } from '../module/token-hud.js'
+import { QuickRollSettings } from '../module/token/hud.js'
```
mjeffw, line 6: We don't have to do this right now, but I would like each module to instantiate its own system settings instead of lumping them all here. That might imply a naming scheme (such as `<module-name>-<setting-name>` to ensure uniqueness. And that might mean a migration between the old schema and the new one. Let me know if you disagree with this or have another approach, or if you just want to keep adding to this catch-all file.

module/actor/actor.js
```diff
@@ -2352,23 +2344,28 @@ export class GurpsActor extends Actor {
...
-    let tokens = this.getActiveTokens(true)
-    for (const token of tokens) {
-      token.setEffectActive(name, active)
+    const effects = set.filter(e => !e.isManeuver)
+
+    const visibility = game.settings.get(Settings.SYSTEM_NAME, Settings.SETTING_MANEUVER_VISIBILITY)
+    if (visibility === 'NoOne') return effects
+
+    if (!game.user?.isGM && !this.isOwner) {
+      if (visibility === 'GMAndOwner') return effects
```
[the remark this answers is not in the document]
mjeffw: I agree with Copilot -- rename to something meaningful.

module/actor/maneuver.js
```diff
@@ -1,5 +1,5 @@
+import { TokenGURPS } from '../token/index.js'
```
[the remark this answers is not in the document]
mjeffw: If Copilot is right, please remove the import.

module/actor/maneuver.js
```diff
@@ -1,5 +1,5 @@
...
 import * as settings from '../../lib/miscellaneous-settings.js'
```
mjeffw: Again, a comment you don't have to fix immediately: Following along with the comments on `global.d.ts`, we should be consistent in our naming. So if we import `* as Settings` in one place, we should do that everywhere and not vary it (such as `* as settings`).

module/combat-tracker/quick-roll-menu.js
```diff
@@ -4,7 +4,8 @@ import { TokenActions } from '../token-actions.js'
 export const addQuickRollButton = async (html, combatant, token) => {
   const quickRollSettings = game.settings.get(Settings.SYSTEM_NAME, Settings.SETTING_USE_QUICK_ROLLS)
   const canShowButtons = quickRollSettings.enabled && (game.user.isGM || combatant.isOwner)
-  if (!canShowButtons || !token.actor) return html
+
```
mjeffw, line 7: I was glad to see this -- I'm a big proponent of using whitespace/blank lines to make code more readable!

module/damage/applydamage.js
```diff
@@ -514,12 +514,11 @@ export default class ApplyDamageDialog extends Application {
...
-          (effect?.modifier ?? 0) === 0
-            ? 'HT'
-            : effect.modifier < 0
-              ? `HT+${-effect.modifier}`
-              : `HT-${effect.modifier}`
+              : `HT-${effect.modifier}`(effect?.modifier ?? 0) === 0
+                ? 'HT'
+                : effect.modifier < 0
+                  ? `HT+${-effect.modifier}`
+                  : `HT-${effect.modifier}`
```
mjeffw: I probably wrote this code -- shame on me. That's too many nested tertiary expressions. I need to break them out into nested if-statements or something else more readable.

module/token/object.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * We use this function because maneuvers are special Active Effects: maneuvers don't apply
+   * outside of combat, and only one maneuver can be active simultaneously. So we really don't
+   * deactivate the old maneuver and then activate the new one -- we simply update the singleton
+   * maneuver data to match the new maneuver's data.
+   */
+  async setManeuver(maneuverId: string): Promise<void> {
+    // if not in combat, do nothing
+    if (!isCombatActive || !isTokenInActiveCombat(this)) return
```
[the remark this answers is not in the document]
mjeffw: @rinickolous Check on this comment; potentially incorrect code.

module/token/object.ts
```diff
@@ -0,0 +1,85 @@
...
+    console.log('TokenGURPS.setManeuver', maneuverId)
+
+    const maneuver = Maneuvers.get(maneuverId)
+    if (!maneuver) return
+
+    const activeManeuvers = Maneuvers.getActiveEffectManeuvers(Array.from(this.actor?.effects.values() ?? []))
+    // if there is a single active effect maneuver, update its data
+    if (activeManeuvers.length === 1) {
+      // @ts-expect-error: waiting for flag type update
+      if (activeManeuvers[0].getFlag('gurps', 'name') !== maneuverId) activeManeuvers[0].update(maneuver)
```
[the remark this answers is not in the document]
mjeffw: Should we be doing an `await` here?

mjeffw (changes requested):

lib/miscellaneous-settings.js
rinickolous: Raised this as separate issue #2262

global.d.ts
rinickolous: I'm glad you brought that up. I think I was going with `<thing>GURPS` as an aesthetic preference but you're absolutely right about consistency. I'll go back to the original naming scheme.

module/actor/maneuver.js
rinickolous: Resolved in latest commit.

module/token/object.ts
rinickolous: Resolved in latest commit.

rinickolous:
Sorry, no idea what happened to the commits here, I think it was just an issue caused by the removal of `module/global-reference.js` on the r18 branch from elsewhere. Should be all good now.

module/damage/applydamage.js
```diff
@@ -514,12 +514,11 @@ export default class ApplyDamageDialog extends Application {
             ? 'HT'
             : effect.modifier < 0
               ? `HT+${-effect.modifier}`
-              : `HT-${effect.modifier}`
-          (effect?.modifier ?? 0) === 0
-            ? 'HT'
-            : effect.modifier < 0
-              ? `HT+${-effect.modifier}`
-              : `HT-${effect.modifier}`
+              : `HT-${effect.modifier}`(effect?.modifier ?? 0) === 0
```
mjeffw: On second look, this code is wrong.  It looks like a copy-paste error -- everything after ```: `HT-${effect.modifier}``` on this line is a repeat of lines 513-517 and should be removed.

mjeffw (changes requested):

module/damage/applydamage.js
rinickolous: Donezo. I thought it looked off.

mjeffw (approved):

PR #2268, opened 2025-05-18
Add yard conversion and speed range penalty methods to RulerGURPSv12

### LINKED ISSUE #2267: V12 Compatibility: ruler.convert_to_yards is not a function
In Foundry v12: Select a target, select another token, and roll an attack from selected token.

```
effect-modifier-popout.js:22 Uncaught (in promise) TypeError: undefined. ruler.convert_to_yards is not a function
[Detected 1 package: system:gurps(0.18.0)]
    at calculateRange (effect-modifier-popout.js:22:25)
    at getRangedModifier (effect-modifier-popout.js:33:15)
    at GurpsActor.addTaggedRollModifiers (actor.js:3067:30)
    at doRoll (dieroll.js:128:42)
    at async Object.attack (gurps.js:863:20)
    at async Object.performAction (gurps.js:1249:15)
    at async Object.executeOTF (gurps.js:336:30)
```

mjeffw:
Closes #2267

### DIFF
--- a/module/canvas/ruler-12.ts
+++ b/module/canvas/ruler-12.ts
@@ -45,4 +45,50 @@ export class RulerGURPSv12 extends Ruler {
     }
     return 0
   }
+
+  convert_to_yards(numeric: number, Unit: string): number {
+    let meter = 0
+    let unit = Unit.toLowerCase()
+
+    if (meters.includes(unit)) meter = numeric
+    else if (millimeters.includes(unit)) meter = numeric / 1000
+    else if (kilometers.includes(unit)) meter = numeric * 1000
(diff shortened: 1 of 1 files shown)


rinickolous (changes requested):
Sorry, didn't realise that something was still using the convert_to_yards function. However, this functionality is effectively duplicating existing functionality which I moved over to Length. Reintroducing `convert_to_yards` can be avoided by instead calling `Length.from(dist, canvas.scene.grid.units).to(Length.Unit.Yard).value`.

Also, the protected method `_yardsToSpeedRangePenalty` is duplicated here and made public. Why not just make the existing function public?

Issue #2288, opened 2025-05-26
Damage shows on roll when using Token Action HUD buttons, not when rolling from the sheet in Foundry 13 GGA 0.18.0-a


TheFabulousIronChef:
Weird observation--when I click on the skill for an attack in the combat tab on the sheet, it rolls against that number but does not show a damage button in the chat window. When I use Token Action HUD, though, it works.

rinickolous:
Oops, looks like I fixed this in #2299 which was intended as part of a separate PR. Oh well, fixed!

PR #2292, opened 2025-05-28
Fixed various bugs in 0.18.0

### LINKED ISSUE #2287: Add to Quick Roll not working for weapon attacks -- cannot click on them in Foundry v. 13 GGA 0.18.0-a
For whatever reason, you can double click on a weapon on the sheet, and add "Add to Quick Roll," but when you do so, the attack portion does not work. The defense (parry or block) interestingly does, but not the attack. 

Image

Image

### LINKED ISSUE #2290: Cannot set maneuvers via token in v. 0.18.0-a
When you click on the maneuver icon, the maneuvers flicker for a nanosecond and disappear before you can select them. Note that the nanosecond is an approximate measure and not actual.

rinickolous:
- Re-worked maneuver selection palette for both Foundry V12 and V13. It's now part of the GurpsTokenHUD subclasses (different subclasses depending on Foundry version). This currently uses the active maneuver image for the menu icon which I think is neat but can be reverted if you don't. Resolves #2290 .
- Fixed damage follow-on not showing for attacks rolled from the character sheet. Resovles #2288 .
- Fixed melee attacks not being clickable in the quick roll menu. Resolves #2287 .
- Fixed the quick roll menu appearing incorrectly in Foundry v13.
- Fixed `SemanticVersion.toString` not outputting the pre-release version and other build data to the resulting string.
- Fixed some modifiers appearing twice in the effect modifier pop-out window.
- Fixed the effect modifier pop-out button not appearing in the canvas controls in Foundry V12.
- Fixed maneuvers not clearing from the token when it is deleted.
- Added forcing tokens to re-draw effects when the Token Maneuvers settings are changed.

EDIT:
I also added native movement cost indicators for the token ruler (for FoundryVTT v13 only) in this commit. Part of that change depends on the previous changes to the `module/token/index.ts` file so I'm bundling it in with this for now but can revert that and submit separately if you wish.
image


mjeffw:
> EDIT:
> I also added native movement cost indicators for the token ruler (for FoundryVTT v13 only) in this commit. Part of that change depends on the previous changes to the `module/token/index.ts` file so I'm bundling it in with this for now but can revert that and submit separately if you wish.

In general I prefer every issue to be a different PR. If they are bundled together, it means rejecting the whole thing if there is something that needs to be fixed with a single change. And it is considered a better development practice -- we prefer several small PRs over a single large one. It makes review easier, too. 

See: 
[pull-request-best-practices](https://graphite.dev/blog/pull-request-best-practices)
[helping-others-review-your-changes](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/getting-started/helping-others-review-your-changes)
[written-unwritten-guide-pull-requests](https://www.atlassian.com/blog/git/written-unwritten-guide-pull-requests)

Also: Please update changelog.md for each GitHub issue (bug or enhancement) that you address as a part of your PRs.

module/token/quick-roll-settings.ts
```diff
@@ -0,0 +1,54 @@
+import { DeepPartial } from 'fvtt-types/utils'
+import * as Settings from '../../lib/miscellaneous-settings.js'
+
+class QuickRollSettings extends foundry.applications.api.HandlebarsApplicationMixin(
+  foundry.applications.api.ApplicationV2
```
mjeffw, line 5: I've got a lot to learn -- what does this mix-in do? Is there some documentation somewhere? The API documentation is not particularly enlightening. I assume this is where `static PARTS` comes from, what does that do?

module/token/token-hud-12.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  /* ---------------------------------------- */
+
+  override activateListeners(html: JQuery<HTMLElement>): void {
+    super.activateListeners(html)
+    this.toggleManeuverTray(this.#maneuverTrayActive)
+    const effectsTray = html.find('.maneuver-palette')
+    effectsTray.on('click', '.maneuver-control', this.#onSetManeuver.bind(this))
+
+    html.find('.control-icon[data-action="effects"]').on('click', () => this.toggleManeuverTray(false))
```
mjeffw, line 48: I assume this closes the maneuver tray once someone clicks on a maneuver.

module/token/quick-roll-settings.ts
rinickolous: ApplicationV2 added separation between the Handlebars application handling and the Application core. This is so that in theory you could even slot in another language like Vue instead of using Handlebars. Foundry's own AppV2 applications extend the Handlebars mixin because they all use Handlebars.

The PARTS static object is a replacement for the template definitions in AppV1. It adds Tue ability to deal with different sections of an application independently (e.g. do partial renders, fetch context date per part, etc.). There's [a community wiki page](https://foundryvtt.wiki/en/development/guides/converting-to-appv2) for this.

module/token/token-hud-12.ts
rinickolous: No, that's handled elsewhere. This closes the maneuver tray if you open the effects tray so they're not both open at the same time and overlapping.

PR #2309, opened 2025-06-01
Bugfix/2285 module conventions

mjeffw:
Let me know if you think this is worthwhile to pursue as a standard for our internal modules.


module/damage/index.ts
```diff
@@ -17,3 +16,6 @@ export function init() {
 }
 
 export { rollDamage } from './roll-damage.js'
+
+// @ts-expect-error
+const _typecheck: GurpsModule = { init }
```
rinickolous, line 21: Rather than declare the type and use a @ts-expect-error comment when not usig them, why not declare the GurpsModule type in the global scope?

module/gurps.js
```diff
@@ -109,11 +109,9 @@ if (!globalThis.GURPS) {
     GURPS.parseDecimalNumber = parseDecimalNumber
   }
 
-  Canvas.init() // Initialize the Canvas module
-  Combat.init() // Initialize the Combat module
-  Damage.init() // Initialize the Damage module
-  Token.init() // Initialize the Token module
-  UI.init() // Initialize the UI module
+  /** @type GurpsModule[] */
+  GURPS.modules = [Canvas, Combat, Damage, Token, UI]
```
rinickolous, line 113: I've been thinking about this approach more, and I'm not really happy with the module names being defined in the import line. Why not make the module name intrinsic to the module rather than defined extrinsically? For example:

```ts
export const Canvas: GurpsModule = {
    init() {
    // ...
    }
}

module/damage/index.ts
mjeffw: Done, although I declared it in its own file (`module/gurps-module.ts`).

module/gurps.js
mjeffw: Agreed

mjeffw: There's an interesting thing that happened when I took your suggestion and explicitly declared the GurpsModule inside the index.ts file: It became a handy place to put any other needed exports, so that instead of importing those things separately, you just import the module and get them all. This adds in the namespace scoping you were talking about some time ago. So for example, instead of exporting the `rollDamage` function directly, I just put it into the module, and a caller would import the module (`Damage`) and access the function via that: `Damage.rollDamage()`.

PR #2312, opened 2025-06-01
Bugfix/tracker editor cannot set breakpoints

mjeffw:
Review this one after reviewing #2318 as it contains some of the same code.

### DIFF
--- a/module/resource-tracker/resource-tracker-editor.js
+++ b/module/resource-tracker/resource-tracker-editor.js
@@ -4,11 +4,10 @@ export class ResourceTrackerEditor extends Application {
    * @param {import('./types.js').TrackerInstance} tracker data to update
    * @param {*} options
    */
-  constructor(tracker, isActor, options = {}) {
+  constructor(tracker, options = {}) {
     super(options)
 
     this._tracker = tracker
-    this._isActor = isActor
   }
 
   /** @override */
@@ -29,7 +28,6 @@ export class ResourceTrackerEditor extends Application {
   getData(options) {
     const data = super.getData(options)
     data.tracker = this._tracker
-    data.isActor = this._isActor
     return data
   }
 


module/damage/index.ts
```diff
@@ -17,3 +16,6 @@ export function init() {
 }
 
 export { rollDamage } from './roll-damage.js'
+
+// @ts-expect-error
+const _typecheck: GurpsModule = { init }
```
rinickolous: What does this do?

module/gurps.js
```diff
@@ -2238,20 +2239,20 @@ if (!globalThis.GURPS) {
...
-    resourceTrackers.forEach(it => (GURPS.DamageTables.damageTypeMap[it.alias] = it.alias))
-    resourceTrackers.forEach(
-      it =>
-        (GURPS.DamageTables.woundModifiers[it.alias] = {
-          multiplier: 1,
-          label: it.name,
-          resource: true,
-        })
-    )
+    // // get all aliases defined in the resource tracker templates and register them as damage types
```
rinickolous: If this is no longer needed then it should be deleted rather than commented out.

module/resource-tracker/index.ts
```diff
@@ -0,0 +1,62 @@
...
+import { OLD_SETTING_TEMPLATES, SETTING_TRACKER_EDITOR, SETTING_TRACKER_TEMPLATES } from './types.js'
+
+export function init() {
+  console.log('GURPS | Initializing GURPS Resource Tracker Module')
+  Hooks.once('ready', async function () {
+    if (!game.settings) throw new Error('GURPS | Game settings not found')
+    if (!game.i18n) throw new Error('GURPS | Game i18n not found')
+
+    // TODO: Remove this when the setting is removed.
+    game.settings.register(GURPS.SYSTEM_NAME, OLD_SETTING_TEMPLATES, {
```
rinickolous, line 20: I've been considering moving `game.settings.register` calls for modules to a `<module name>/settings.ts` file rather than having them in the index, just to make the index cleaner. What do you think?

module/resource-tracker/types.ts
```diff
@@ -0,0 +1,38 @@
+// TODO New setting names:
+// - SETTING_TRACKER_DEFAULT_EDITOR ('tracker-manager') becomes 'tracker.manager'
+// - SETTING_TRACKER_TEMPLATES ('tracker-templates') becomes 'tracker.templates'
+
+// README: By Convention, settings should be prefixed with the module name.
+const SettingPrefix = 'resource-tracker'
```
rinickolous, line 6: I like this, very clean.

module/resource-tracker/resource-tracker-manager.js
```diff
@@ -1,27 +1,10 @@
-import * as Settings from '../../lib/miscellaneous-settings.js'
 import { arrayToObject, objectToArray } from '../../lib/utilities.js'
-import { ResourceTrackerEditor } from './resource-tracker-editor.js'
```
rinickolous, line 3: `ResourceTrackerEditor` is now failing to resolve from within the resource tracker manager settings application.

templates/resource-tracker/tracker-manager.hbs
```diff
@@ -0,0 +1,54 @@
...
+              type='text'
+              readonly
+              value='{{tracker.name}}'
+              placeholder='{{localize "GURPS.clickToEdit"}}'
+            />
+            <!-- tracker.alias -->
+            <input name='alias' data='{{@index}}' class='gcs-input' type='text' readonly value='{{tracker.alias}}' />
+
+            <!-- slot -->
+            <input name='autoapply' name='slot' type='checkbox' data='{{@index}}' {{checked slot}} />
```
rinickolous: the autoapply value is failing to save. You have two name properties defined here, I think one of them is a mistake.

rinickolous (changes requested):
A few issues to resolve and then just some questions.

module/damage/index.ts
mjeffw: Simply verifies that the object matches the shape of the interface. Not needed anymore as I moved to an explicit declaration and export.

module/resource-tracker/index.ts
mjeffw: For many modules, this is the only setup required, so I'm good with leaving them in the init() function.

module/resource-tracker/resource-tracker-manager.js
mjeffw: Fixed.

module/damage/damagechat.js
```diff
@@ -91,10 +91,10 @@ export default class DamageChat {
 
   /**
    * Create the damage chat message.
-   * @param {GurpsActor|User} actor that rolled the damage.
+   * @param {Actor<"base" | ModuleSubtype>|User} actor that rolled the damage.
```
rinickolous, line 94: I believe this should be Actor.Implementation instead

rinickolous (approved):
Just one non-urgent suggestion on typing. I'm not able to test properly as am not at my computer this weekend but everything seems fine. Happy to merge if you are.

module/damage/damagechat.js
mjeffw: I get TS compilation errors when I make that change. I'll leave it for now and we can resolve it in the next release.

PR #2315, opened 2025-06-03
Bugfix - #2314 - Updating an actor creates a new instance of the rendered actor sheet

### LINKED ISSUE #2314: Updating an actor creates a new instance of the rendered actor sheet
This seems to only happen in FoundryVTT version 12. To re-create: Open the editor sheet and edit any field, or even try to close a sheet (which updates the actor).

rinickolous:
This option is really only used for forcing an application to re-render afaik and shouldn't be used this casually. We should instead not be updating the actor multiple times with one change and going through data preparation every time. This is bad practice and is insanely slow, yet is very prevalent in the system as it stands. I don't think this will cause any missed renders.

This resolves #2314 which was affecting users in FoundryVTT v12 (The issue does not seem to affect v13 in the same way).

### DIFF
--- a/module/actor/actor.js
+++ b/module/actor/actor.js
@@ -1513,7 +1513,7 @@ export class GurpsActor extends Actor {
 
   _forceRender() {
     this.ignoreRender = false
-    this.render(true)
+    this.render()
   }
 
   // Drag and drop from an equipment list


mjeffw:
> This is bad practice and is insanely slow, yet is very prevalent in the system as it stands. I don't think this will cause any missed renders.

Be careful about using terms like "insanely slow" -- that seems very subjective and dismissive, and the reality is that performance is currently acceptable.

PR #2318, opened 2025-06-06
Bugfix/2285 resource tracker templates

mjeffw:
I don't know what is going on with my git branches -- my local bugfix/2285-module-conventions is now pointing to bug fix/2285-resource-tracker-cannot-set-breakpoints on the remote. 

This PR replaces the old one.


module/utilities/i18nFallback.ts
rinickolous, line 1: I'm not at my computer so can't test this, but I would imagine the point of i18nFallback is to provide a fallback string + warning if the i18n utility is not initialised yet.
Instead, making the function throw an error (rather than just log the error) would result in early return and no value returned.

mjeffw: The current code throws an exception if game.i18n doesn't exist, which should be never. The check removes the need to do `game.i18n?.localize` or do an explicit if check. The fallback is needed when some of the values being translated are standardized and localized, and others are user entered (which does not require localizing).

PR #2320, opened 2025-06-09
A set of related changes:

### LINKED ISSUE #2308: BUG: Setting PDF Page Offset and PDF Book Code in new PDF Journals is not possible under v13
GGA Version 0.18.0

FVTT v13 Stable build 334

Back in v12, if you added a PDF to a Journal, you could specify the PDF Source, a Page Offset (to work out how many pages in Page 1 actually started) and PDF Book Code so you could reference it as an internal hyperlink:

Image

 In v13, these options appear to have disappeared:

Image

PDF journals created previously in a world (around FVTT v10, from memory) and upgraded to v13 still work exactly as expected with Page Offset and PDF Book Code all working.  0.18.0 still works as expected in FVTT v12 Stable build 343 (final release).

This suggests it's a UI issue, and that the underlying data structures are still OK.

I don't get a specific console error, but when accessing the "edit" function of the Journal of an existing item, I get the following:

```
grab_to_pan.js:20 [fluent] Missing translations in en-gb: pdfjs-editor-highlight-button, pdfjs-editor-highlight-button-label
formatWithFallback @ localization.js:57
await in formatWithFallback
formatMessages @ localization.js:98
translateElements @ dom_localization.js:274
translateFragment @ dom_localization.js:254
(anonymous) @ dom_localization.js:172
translateRoots @ dom_localization.js:172
translate @ grab_to_pan.js:20
initialize @ app.js:255
await in initialize
run @ app.js:616
webViewerLoad @ viewer.js:40
(anonymous) @ viewer.js:75
__webpack_require__.a @ xfa_layer_builder.js:79
6400 @ ui_utils.js:811
__webpack_require__ @ viewer.js:204
(anonymous) @ bootstrap:2
app.js:1506 PDF 381a8136b82675e57acf8afab0c5feca [1.6 Acrobat Distiller 6.0.1 for Macintosh / QuarkXPress: pictwpstops filter 1.0] (PDF.js: 4.0.379 [9e14d04fd])
```

Not sure if that helps?

At the moment, it's not possible to create a referenceable PDF in v13 as it was in v12

mjeffw:
- Create a PDF module.
- Move PDF settings into the module and ancapsulate them in a Settings Menu.
- Move handlePdf and handleOnPdf to PDF module.
- Convert GURPS module from array to object keyed by module name.
- Fix missing i18n tags for Effect Modifier settings.
- Add methods to get PDF system settings from PDF module.
- Refactor GurpsPDFSheet into a v13 version and a pre-V13 version.
- Refactor pdf/edit.hbs and pdf/view.hbs to work for both v13 and pre-v13 versions.

fixes #2308


global.d.ts
```diff
@@ -5,7 +5,7 @@ import { GurpsCombatant } from './module/combat/combatant.ts'
 import { GurpsItem } from './module/item.js'
 import { GurpsToken } from './module/token/gurps-token.ts'
 
-export { }
+export {  }
```
rinickolous, line 8: not an issue, but why is there seemingly another space added here with every PR?

module/gurps.js
```diff
@@ -2581,15 +2562,18 @@ if (!globalThis.GURPS) {
       GurpsWiring.hookupAllEvents(html)
     })
 
+    // TODO: Move to the combat module?  We could have a method in combat that allows other modules to request hooks into the Combat system.
```
rinickolous, line 2553: agreed on the comment here. we shouldn't really be using hooks for things we can use subclasses for.

templates/pdf/edit.hbs
```diff
@@ -1,31 +1,49 @@
...
-		<label>{{localize "GURPS.pdfCode"}}</label>
-		<div class="form-fields">
-			<input type="text" name="system.code" placeholder="PDF Code (e.g. 'B')" value="{{document.system.code}}">
-		</div>
-	</div>
+  <div class='form-group'>
+    <label>{{localize 'JOURNALENTRYPAGE.PDFSource'}}</label>
+    <div class='form-fields'>
+      <file-picker type='text' name='src' accept='application/pdf, image/*' class='file-picker' value='{{document.src}}'>
+        {{!-- <input class='image' type='text' name='src' placeholder='path/to/file.pdf' value='{{document.src}}' /> --}}
```
rinickolous: Looks like unnecessary commented out elements

module/pdf/sheet.js
```diff
@@ -1,42 +1,78 @@
-export class PDFViewerSheet extends DocumentSheet {
-  pageNumber = 1
+// export class GurpsPDFEditSheet extends foundry.applications.sheets.journal.JournalEntryPagePDFSheet {
+// COMPATIBILITY: v12
 
-  constructor(object, options = { pageNumber: 1 }) {
-    super(object, options)
-    this.pageNumber = options.pageNumber
+export class GurpsPDFSheet extends foundry.applications.sheets.journal.JournalEntryPagePDFSheet {
```
rinickolous: Calling this namespace caused the system to fail to load in v12. To fix this, encapsulate the subclass in a function and return early if the game version is not v13 or newer. See my v13-specific classes for an example.

rinickolous (changes requested):
There's currently an issue with the PDF viewer which occurs when the window is resized:
image

This might be fixed by setting `article.div:first_child()` height to 100% (or something like that) and would be useful to include in this PR.

Other than that, the PR currently doesn't work at all in v12 due to calling an undefined namespace. There are also some other minor comments.

global.d.ts
mjeffw: No idea. Something with my prettier config?

module/pdf/sheet.js
```diff
@@ -1,42 +1,104 @@
...
+    }
   }
+  return GurpsPDFSheet
+}
+
+function getGurpsPDFSheetV1() {
+  if (!game.release) return
+  if (game.release.generation >= 13) return
+
+  class GurpsPDFSheetV1 extends JournalPDFPageSheet {
```
rinickolous, line 64: The v12 PDF sheet is stuck on a specific height because it inherits the default height option "auto" from its parent class. We can prevent this by defining a height. I used a 1.3 aspect ratio by default because it matches the source books I checked.

```suggestion
  class GurpsPDFSheetV1 extends JournalPDFPageSheet {
    static get defaultOptions() {
      return foundry.utils.mergeObject(super.defaultOptions, {
        width: 600,
        height: 780,
      })
    }
```

We can also force the aspect ratio to remain at 1:1.3 by adding to the stylesheet. Here, I am adding another class to the root element to distinguish between the view and edit sheets. We don't want to apply this forced aspect ratio to the edit sheet.

We might not want to do this in case users use other aspect ratios. The above suggestion should be sufficient to resolve current issues.

```suggestion
  class GurpsPDFSheetV1 extends JournalPDFPageSheet {
    activateListeners(html) {
      super.activateListeners(html)

      this.element[0]?.classList.add(this.isView ? 'view' : 'edit')
    }

```

styles/_pdf.scss
```diff
@@ -0,0 +1,15 @@
...
+    overflow: hidden;
+
+    div:first-child {
+      flex-grow: 1;
+      display: flex;
+      flex-direction: column;
+      overflow: auto;
+    }
+  }
+}
```
rinickolous, line 19: As part of the above comment in the sheet definition class, this suggestion should force the aspect ratio to remain at 1:1.3, whatever the actual size of the sheet.

```suggestion
}

.journal-entry-page.pdf.view {
  aspect-ratio: 1 / 1.3; // 1 to 1.3 ratio is about right for GURPS source books
}
```

PR #2328, opened 2025-06-21
Migration of GurpsActor and GurpsItem to DataModel, and direct GCA 5 file import

rinickolous:
This is a draft, the functionality is not complete and merging would result in issues at this stage.

> [!WARNING]
> This PR should at this stage be tested on a new world (or after taking a backup of an existing world) to avoid data loss. This PR replaces the existing GurpsActor and GurpsItem classes with new classes with a slightly different data structure.

This PR includes the following:
- A full conversion of the data structure of GURPS Actors and Items to use Foundry's DataModel
- Full type-safe imported data schemas for both GCS 5.36 and GCA 5
- Full import functionality from the aforementioned programs to the aforementioned DataModels

To test the new import functionality, run the command `GURPS.importGCS()` or `GURPS.importGCA()`. Please note that the resulting Actor does not display correctly on the sheet as updating sheets to match the slight differences in the schema are currently beyond the scope of this PR. You can inspect the output of the import function in the console at this stage.

While this is a large PR, I believe it sort of obeys the rule of being limited in scope to one thing.

I am raising this PR at this stage because the work contained within the scope of "convert Actors and Items to use DataMoel and make sure importing still works" is almost entirely complete, and I don't want to add anything else in at this stage in case I've gone off the trail somewhere.

The data schema of the actor system object is slightly different than the original. All changes to the existing schema have been marked with a `// NOTE: ` comment explaining the change. The schema changes follow 3 general types:
- Unused fields: These are fields which are not referred to anywhere in the code. They have either been defined and commented out, or mentioned by name but not defined.
- Dynamic fields: These are fields representing values wich are only available at runtime, following data preparation. They have been moved to class instance properties defined during data preparation.
- Type-changed fields: These are fields which previously had a type which was suboptimal for the field's function. These include ArrayFields which were previously pseudo-arrays (objects using zero-filled integers as keys) and some instances of inappropriate primitive types (e.g. the field representing Size Modifier was previously a string, but makes much more sense as a number)

Current potential issues include interactions with the ADD. The new importers directly import the hit location table from both programs rather than use the standard set of hit locations defined within GGA. If we can figure out some way of doing so, I think we should at this stage move to allowing arbitrary hit location tables.

Please suggest any changes to my current changes, review for any obvious issues, and let me know if I am clear to continue. My plan going forward is the following:
- Convert existing Actor Sheets to correctly display and interact with the new structure. I belive the cleanest way of doing this at this stage is to convert functions from `GurpsActor` to functions of the same name in `GurpsActorV2` and simply change the functionality to work with the new model. This minimises the need to change files outside of the Actor Document and DataModel files themselves, though may still be necessary in some cases.
- Create functionality to migrate Actors using the previous schema to the new one.
- Convert existing non-sheet functionality to interact with GurpsActorV2. Once again, the plan is largely to simply convert functions from GurpsActor to avoid unnecessarily expanding the file scope of this change.

A breakdown of changes by file follows:

- TODO.md: This is a working document with some analysis of the current state of the Actor data schema and migration notes. It can be ommitted when no longer needed, but useful for now.
- _template.json: This has been renamed so Foundry no longer recognises it. Defining a template.json is not necessary following the introduction of DataModels.
- dev-utilities/ files: These are reference files for the native structure of GCA5 files.
- global.d.ts: Defined more settings and some classes used
- module/action files: As previously discussed, weapons should not be items. Borrowing from a few other systems (dnd5e, draw-steel), I am introducing the Pseudo-Document which is extended in the Action class (and subclasses MeleeAttack and RangedAttack) to give some Document-like functionality in what is otherwise just a DataModel.
- module/actor/data files: This is the newly defined `CharacterModel` which aims to as closely as reasonable copy the previous data structure. The model is not entirely complete (some functionality from GurpsActor is missing, but the plan is to add it before merging). Some component schemas are located in a separate file for cleanliness.
The actor contains some data preparation logic refactored from GurpsActor, though this is largely untested at this stage.
- module/item/data files: These files define the structure of the system objects in GurpsItemV2. They include some data preparation methods used to offload the applying of bonuses from the actor class itself. These, like the above, are largely untested at this stage but will be tested before merging.
- module/utilities/gcs-importer/: This tool defines a clear structure for importing GCS 5.36 files directly, and offers the ability to import them to the GGA Actor schema in a type-safe way, using Foundry's DataModel for data validation and type resolution when importing files.
- - module/utilities/gca-importer/: This tool defines a clear structure for importing GCA 5 files directly, and offers the ability to import them to the GGA Actor schema in a type-safe way, using Foundry's DataModel for data validation and type resolution when importing files.
- system.json: As a replacement for template.json, system.json is now the place to define parts of your Document structure. This is limited to HTML and filepath fields as these require server-side processing. Since our Documents don't use these, the JSON objects corresponding to them are empty.


mjeffw:
Can you close this PR and open a new one to merge into `develop/v13-only` (instead of `main`)?

PR #2336, opened 2025-07-02
Fix item cost and weight calculations to use 'calc' properties

### LINKED ISSUE #2327: Some fields are not imported when importing a GCS equipment library
- Foundry version: 12.343 and 13.345
- GURPS Game Aid version: 0.18.1
- GCS version: 5.36.1

When importing a GCS equipment library as a compendium, the following fields are failing to import:

- Cost
- Weight
- Notes

The fields used to import but are no longer importing - I'm not sure what the last version was when they imported correctly. When "Import Extended Values from GCS Compendiums" is selected in the settings, the extended cost & weight values are correctly imported, which has served as a work-around for items without equipment modifiers. Items with equipment modifiers that modify cost/weight still have to be manually updated as the extended cost/weight in the GCS library will include all optional equipment modifiers. Equipment modifiers are not imported, but I don't remember if that was functionality that worked before and is now broken or if it would be a new feature.

Zip file contains an example GCS equipment library that when imported is showing the issue.
[Equipment.eqp.zip](https://github.com/user-attachments/files/20848641/Equipment.eqp.zip)

mjeffw:
Fixes #2327

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,11 @@
 ### [Users Guide](https://bit.ly/2JaSlQd) for GURPS 4e Game Aid for Foundry VTT
 
+## Release 0.18.2
+
+### Bugfixes
+
+- Some fields are not imported when importing a GCS equipment library #2327
+
 ## Release 0.18.1 06/21/2025
 
 ### Features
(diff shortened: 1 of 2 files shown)


rinickolous (approved):
No issues that I can see. Merging.

PR #2400, opened 2025-11-27
Made new Actor import functionality accessible from character sheet.

rinickolous:
- Added optional actor paramter to importGCS function for importing data on top of existing Actor Document.
- Actor import function now checks for Actor type and displays appropriate import dialog depending on type.
- Added localization for all existing Item and Actor types.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -710,6 +710,7 @@
     "importSelectFileOverwriteAction": "Overwrite the data for: <b>{name}</b>.",
     "importSelectFileSource": "Source Data",
     "importSelectFileTitle": "Select a file exported from GCS or GCA.",
+    "importSelectFileTitleGCS": "Select a file exported from GCS 5+.",
     "importSheetHint": "Importing data for {name} using {generator}. Please wait...",
     "importSheetTitle": "Importing {generator} Sheet",
     "importSource": "Source",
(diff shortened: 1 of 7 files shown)


mjeffw:
Is there any reason to present "Enemy" as a character type to the user? If we want to implement something in the future, we can add it back.

module/gcs-importer/gcs-importer.ts
```diff
@@ -1,7 +1,7 @@
 import { GcsImporter } from './importer.js'
 import { GcsCharacter } from './schema/character.js'
 
-async function importGCS() {
+async function importGCS(actor?: Actor.OfType<'characterV2'>) {
   return new foundry.applications.api.DialogV2({
     window: {
       title: 'Import from GCS 5.36',
```
mjeffw: I know this is a line that hasn't changed -- but I don't think we should be so restrictive as to only allow GCS 5.36. I would prefer to allow any GCS v5+ version.

lang/en.json
```diff
@@ -1,4 +1,21 @@
 {
+  "TYPES": {
+    "Actor": {
+      "character": "Character (v1)",
+      "characterV2": "Character (v2)",
+      "enemy": "Enemy"
```
mjeffw, line 6: Let's remove enemy from the list. It will only confuse folks.

rinickolous:
> Is there any reason to present "Enemy" as a character type to the user? If we want to implement something in the future, we can add it back.

Enemy will show up either way considering it's in the `system.json` file. Are you suggesting that we remove it? I wouldn't be against it but for the fact that some users may be using `enemy` actor types which may be read as invalid by such a change.

PR #2405, opened 2025-12-09
Added HP & FP override prompt to new GCS importer

rinickolous:


module/gcs-importer/importer.ts
```diff
@@ -223,6 +228,66 @@ class GcsImporter {
...
+      })
+
+      if (stopOverwrite) {
+        this.output.HP!.value = currentHP
+        this.output.FP!.value = currentFP
+      }
+
+      return
+    }
   }
```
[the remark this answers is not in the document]
rinickolous: no stat difference is already accounted for, and setting validation is handled elsewhere. no need to check for it here.

lang/en.json
mjeffw, line 1: Let's keep the camelCase for language tags in these files.

lang/en.json
```diff
@@ -1583,6 +1583,34 @@
...
-    "youDoNotHavePermssion": "You do not have permission to do that."
+    "youDoNotHavePermssion": "You do not have permission to do that.",
+    "Dialog": {
+      "Save": "Save",
+      "Overwrite": "Overwrite"
+    },
+    "Importer": {
+      "Prompt": {
+        "Title": "Import Character Data for {name}",
+        "Description": "Select a file exported from GCS 5+.",
```
mjeffw: We no longer require the file to be exported from GCS; this should probably read something like "Select a GCS 5+ character file".

lang/en.json
```diff
@@ -1583,6 +1583,34 @@
...
+        "Source": "Source Data",
+        "Note": "This import will overwrite the data for: <strong>{name}</strong>.",
+        "Warning": "<span class='fas fa-exclamation-triangle'></span>&nbsp;<strong>Note</strong>: This action cannot be undone!"
+      },
+      "PromptHPandFP": {
```
mjeffw: If the setting is the same as one of the legacy settings, remove the old language file entries.

module/gcs-importer/importer.ts
```diff
@@ -223,6 +228,64 @@ class GcsImporter {
...
+        fp: `${this.output.FP!.value}`,
+      }),
+      modal: true,
+      buttons: [
+        {
+          action: 'save',
+          label: game.i18n!.localize('GURPS.Dialog.Save'),
+          icon: 'far fa-square',
+          default: true,
+          callback: () => true,
```
mjeffw: I prefer not using the callback for this type of dialog. With no callback, it returns the action text ('save' or 'overwrite'). I do something similar in `GurpsActorV2.moveEquipment`.

mjeffw: Also, just a general statement about DialogV2 -- I prefer NOT doing any work inside the Dialog, but instead just returning the user input. That may not always be the simplest approach, in which case we can make a call at that time.

module/gcs-importer/importer.ts
```diff
@@ -223,6 +228,65 @@ class GcsImporter {
...
+  }
+
+  /* ---------------------------------------- */
+
+  async #promptPointPoolOverwrite() {
+    if (!this.actor) return // No need to run this if there is no existing actor
+    const currentHP = this.actor.system.HP.value
+    const currentFP = this.actor.system.FP.value
+
+    const statsDifference = currentHP !== this.output.HP!.value || currentFP !== this.output.FP!.value
```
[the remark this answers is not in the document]
rinickolous: currentHP is forced to be a number due to DataModel type enforcement.

module/gcs-importer/settings.ts
```diff
@@ -0,0 +1,70 @@
...
+      choices: {
+        yes: game.i18n.localize('GURPS.settingImportHPAndFPUseFile'),
+        no: game.i18n.localize('GURPS.settingImportHPAndFPIgnore'),
+        ask: game.i18n.localize('GURPS.settingImportHPAndFPAsk'),
+      },
+      initial: 'ask',
+    }),
+    onChange: value => {
+      // Old setting no longer shows up so set it through this one.
+      const oldValue = value === 'yes' ? 0 : value === 'no' ? 1 : 2
```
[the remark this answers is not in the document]
rinickolous: Incorrect.

PR #2406, opened 2025-12-10
Revert "Fixed: GurpsSettingsApplication does not respect StringField …

mjeffw:
…choices"


mjeffw:
Your changes broke my CSS formatting of the GurpsSettingsApplication. Before your changes, the settings dialog looked like this:
image

After your changes, it looks like this:
image

PR #2415, opened 2025-12-12
Add Body Plan overwrite prompt to new importer

rinickolous:


module/actor/actor-importer.js
```diff
@@ -1322,10 +1322,10 @@ export class ActorImporter {
 
   async getSaveOrOverwriteBodyPlan(saveprot, currentPlan, newPlan) {
```
mjeffw, line 1323: I know this PR didn't change it, and I know I probably wrote this, but the method name is `getSaveOrOverwriteBodyPlan`, but it returns `true` or `false` which isn't consistent. 

Either the method name should be changed to something like `isBodyPlanSaved` or the return value should be (`'save'` or `'overwrite'`) and the callers updated accordingly.

I prefer the latter.

module/gcs-importer/importer.ts
```diff
@@ -317,8 +317,8 @@ class GcsImporter {
 
       // Try to determine the role of the hit location. This is used in the Damage Calculator to determine crippling
       // damage and other effects.
-      let temp = hitlocationDictionary![this.output.additionalresources!.bodyplan!.toLowerCase()] as any
-      let entry = Object.values(temp).find((entry: any) => entry.id === location.id)
+      let tempEntry = hitlocationDictionary![this.output.additionalresources!.bodyplan!.toLowerCase()] as any
```
mjeffw, line 320: Good change on this variable name.

module/gcs-importer/importer.ts
```diff
@@ -330,6 +330,74 @@ class GcsImporter {
...
+      delete location._dr
+      delete location.drCap
+      delete location.drItem
+      delete location.drMod
+      return location
+    })
+
+    const statsDifference =
+      currentBodyPlan !== this.output.additionalresources!.bodyplan ||
+      !(this.output.hitlocationsV2 as any[]).equals(currentHitLocations)
```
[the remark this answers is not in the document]
mjeffw: I agree with this comment.

module/gcs-importer/importer.ts
```diff
@@ -333,6 +333,74 @@ class GcsImporter {
...
+
+    if (!statsDifference) return
+
+    const promptSetting = game.settings?.get(GURPS.SYSTEM_NAME, OVERWRITE_BODYPLAN)
```
mjeffw: Rename `promptSetting` to something with semantic meaning, such as `automaticOverwrite`.

module/gcs-importer/settings.ts
```diff
@@ -2,9 +2,9 @@ import { GurpsSettingsApplication } from '../utilities/gurps-settings-applicatio
 import fields = foundry.data.fields
 
 // @deprecated constant
-import { SETTING_IMPORT_HP_FP } from '../../lib/miscellaneous-settings.js'
+import { SETTING_IMPORT_HP_FP, SETTING_IMPORT_BODYPLAN } from '../../lib/miscellaneous-settings.js'
```
mjeffw: Why import this instead of defining it here in the gcs-importer module?

module/gcs-importer/settings.ts
```diff
@@ -26,9 +26,27 @@ export default function initializeGameSettings() {
...
+    config: true,
+    default: 2,
+    type: Number,
+    // @ts-expect-error: weird type nonsense
+    choices: {
+      0: game.i18n.localize('GURPS.importer.settings.overwriteBodyPlan.yes'), // Yes, always overwrite
+      1: game.i18n.localize('GURPS.importer.settings.overwriteBodyPlan.no'), // No, never overwrite
+      2: game.i18n.localize('GURPS.importer.settings.overwriteBodyPlan.ask'), // Ask before overwriting
+    },
+  })
```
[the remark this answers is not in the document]
mjeffw: I agree with this comment.

mjeffw (changes requested):

module/gcs-importer/importer.ts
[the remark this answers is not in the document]
rinickolous: Foundry adds this functionality to Arrays. It's in their [API documentation](https://foundryvtt.com/api/functions/primitives.Array.equals.html).

module/gcs-importer/settings.ts
```diff
@@ -52,6 +70,30 @@ export default function initializeGameSettings() {
...
+    type: new fields.StringField({
+      required: true,
+      nullable: false,
+      choices: {
+        yes: game.i18n.localize('GURPS.importer.settings.overwriteBodyPlan.yes'), // Yes, always overwrite
+        no: game.i18n.localize('GURPS.importer.settings.overwriteBodyPlan.no'), // No, never overwrite
+        ask: game.i18n.localize('GURPS.importer.settings.overwriteBodyPlan.ask'), // Ask before overwriting
+      },
+      initial: 'ask',
+    }),
```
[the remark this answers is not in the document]
rinickolous: I don't think this is accurate for settings which use a DataModel field as the type.

module/gcs-importer/settings.ts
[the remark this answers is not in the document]
rinickolous: Corrected.

module/actor/actor-importer.js
rinickolous: Corrected.

module/gcs-importer/settings.ts
rinickolous: I wanted to avoid declaration duplication and/or changing imports in the "legacy" files. Would you prefer that I duplicate the constant declarations, or move them and then change where they're being imported from in the legacy files?

module/gcs-importer/importer.ts
rinickolous: Done.

module/gcs-importer/settings.ts
mjeffw: Move them.

module/gcs-importer/types.ts
```diff
@@ -1,3 +1,9 @@
 export const MODULE_NAME = 'importer'
 
+// @deprecated: old setting name
+export const SETTING_IMPORT_HP_FP = 'import_hp_fp'
+// @deprecated: old setting name
+export const SETTING_IMPORT_BODYPLAN = 'import_bodyplan'
```
mjeffw, line 6: Move these to the new file and update references as mentioned in the conversation thread.

module/gcs-importer/settings.ts
mjeffw: Actually, the way I was doing our "submodules" is that only the submodule knows about the settings and presents a method to return the current setting value in the module interface.

Example: `module/pdf/index.ts`

PR #2428, opened 2025-12-17
Stopped blank portraits from being imported from GCS

rinickolous:


module/gcs-importer/importer.ts
```diff
@@ -162,9 +162,20 @@ class GcsImporter {
       return
     }
 
-    if (game.user?.hasPermission('FILES_UPLOAD')) {
-      this.img = `data:image/png;base64,${this.input.profile.portrait}.png`
+    if (!game.user?.hasPermission('FILES_UPLOAD')) {
+      console.warn(
```
mjeffw: Should this also be displayed as a UI notification?

rinickolous: Probably, sure. I'll amend when I get home.

rinickolous: Probably, sure. I'll amend when I get home.

PR #2433, opened 2025-12-19
Bugfix/2423 changing posture from character sheet takes two tries

mjeffw:


lib/miscellaneous-settings.js
rinickolous, line 1: Changes to this file (and the lang files) don't seem like they're in scope for this bugfix PR. Can they be removed? I think they're already part of your other (feature) PR anyway.

mjeffw: Yeah, sorry about that. I branched from the wrong source branch. Since they are the same in both branches I am going to leave them here.

PR #2452, opened 2025-12-30
fix: prevent crash when rolling from actor sheet with canvas disabled

### LINKED ISSUE #2449: [Bug?] Cant roll from Actor Sheet when Canval disabled
If I’m “playing” with the canvas disabled, then when I try to make a roll from the character sheet nothing happens (the roll dialog doesn’t appear).

At the same time, the console shows the error:
Uncaught (in promise) TypeError: can't access property "controlled", canvas.tokens is undefined
with a reference to:
doRoll dieroll.js:104

That line contains:
token = actor?.getActiveTokens()?.[0] || canvas.tokens.controlled[0];

It’s clear that with the canvas disabled, this will throw an error.
I noticed that in dieroll.js the canvas is referenced two more times:
const actorTokens = canvas.tokens?.placeables.filter
const actorToken = canvas.tokens?.placeables.find

In both cases there’s protection against canvas.tokens being missing.

Maybe it makes sense to rewrite that line as:
token = actor?.getActiveTokens()?.[0] || canvas.tokens?.controlled[0];

At least in this form, dice rolls from the character sheet started working, and the absence of a token itself doesn’t seem critical.

daynin:
Extract token retrieval logic into getTokenForActor() helper to ensure
  safe access to canvas.tokens when canvas is disabled.

  Closes #2449

### DIFF
--- a/module/dierolls/dieroll.js
+++ b/module/dierolls/dieroll.js
@@ -1,5 +1,6 @@
 import * as Settings from '../../lib/miscellaneous-settings.js'
 import { TokenActions } from '../token-actions.js'
+import { getTokenForActor } from '../utilities/token.js'
 
 export const rollData = target => {
   let targetColor, rollChance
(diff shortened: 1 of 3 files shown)


rinickolous (approved):
Thanks for the fix! Approved.

PR #2454, opened 2025-12-30
fix: add name property to new Active Effect creation

### LINKED ISSUE #2447: [Bug] Cannot create Active Effects from Actor Sheet
Does not seem to matter if it is the Tabbed Sheet or the Full Sheet. 
Gives `GURPSActiveEffect validation errors: name: may not be undefined [Detected 1 package: system: gurps(0.18.10)]`

Image

Image

https://github.com/user-attachments/assets/348ac3a6-146f-404c-a68a-e79e1441b676

daynin:
Foundry VTT v12+ requires the name property for Active Effect validation. Without it, users get "name: may not be undefined" error.

Closes #2447


rinickolous (approved):
Thanks for the fix! Approved.

PR #2456, opened 2025-12-31
Feature/modern UI

mjeffw:
Another patch from daynin.

@rinickolous What's your thoughts on including this in 0.18.x? I love the look of this sheet but I am also concerned about the number of sheets in the base software. (On the other hand, this supports dark mode!)

### DIFF
--- a/styles/modern/components/_notes.scss
+++ b/styles/modern/components/_notes.scss
@@ -26,6 +26,7 @@
 
 .ms-notes-list {
   font-size: var(--ms-font-xs);
+  min-width: 0;
 }
 
 .ms-note-row {
@@ -41,6 +42,8 @@
 
 .ms-note-content {
   flex: 1;
+  min-width: 0;
+  overflow-wrap: break-word;
 
   .expandcollapseicon {
     margin-right: var(--ms-space-2);
--- a/styles/modern/components/_sections.scss
+++ b/styles/modern/components/_sections.scss
@@ -4,6 +4,7 @@
   border-radius: var(--ms-radius);
   padding: var(--ms-gap);
   margin-bottom: var(--ms-gap);
+  min-width: 0;
 }
 
 .ms-section-header {


rinickolous (commented):
I can see some issues with this at the moment:
- The sheet does not respect the Foundry Item data setting, and opens the old non-Document edit dialog either way.
- The HP and FP bars are _ok_, but more appropriate for a system which does not support negative HP/FP values. HP/FP will appear to be depleted at 0 or below. I would rather have different colors reflect the states than have an empty bar for any value below 0.
- The same goes for the portrait gradually filling with a red overlay. It seems to imply that 0 HP is some hard limit whereas this is not the case per the GURPS rules.
- I think some of the way the code is organized is not how I would do it, but this can be fixed without much issue post-merge.
- Some of the types/interfaces added directly to `global.d.ts` may be better placed elsewhere, though to be fair we haven't got much precedent for it.

The sheet styling is quite nice. The design principles in place as per the README are good.
The style sheets use both CSS var variables and SASS variables, which is odd.

The `form-field.hbs` partial appears to be vestigial, it isn't actually used anywhere.

The positive/negative character point summaries under the portrait don't really make sense to me.

---

I would hold off on merging this until after `develop/v13-only` is released. It's a very solid implementation overall and somewhat represents the direction I would like to take the sheets. It's also worth noting that this is still an AppV1 implementation. I would be open to taking elements of this and adding them to an eventual AppV2 implementation.

rinickolous:
> I kind of liked the idea, that if this didn't require much in the way of change, that we could put it into 0.18.x for those users that really want a dark mode sheet before 1.0 is complete.

Yeah, I like that idea too.

> How do you think we should proceed? Thank daynin for the input, give him this feedback, and ask him to resubmit once 1.0 is available? Or before?

On second thought, I'm happy to merge this once the issues above are addressed, and don't have any qualms about doing so before 1.0 is complete. If it means having to convert some stuff over for AppV2, then we would have to do that either way. Happy to include this in 0.18.X

> He's asked to be involved in the transition to TypeScript, fwiw

Excellent!

mjeffw:
> I don't really have an issue with adding this to the base system, though I definitely hear you on the number of sheets being too large at the moment. I think the solution to this is not excluding a sheet like this, but removing some of the other sheets we already have. I don't think we really need the Combat, Inventory Only, or Simple sheets. The Editor sheet shouldn't exist as a separate sheet either. At most, I think that 1-2 sheets for regular characters would be best. Those being: This modern one, and the Tabbed GCS sheet. The Inventory Only and NPC sheets may be somewhat useful, though I wouldn't make the Inventory Only sheet available for characters.

I'd like to move some of the specialty sheets to a module that users can add as needed.

daynin:
> As a GM and a Player, having green and red numbers near the character portrait immediately make me think about HP or FP or ER and so on. I don't need to know the character point total/disad total. As a GM, all my encounters are planned, and as a player, I know my totals already. It's just a distraction, sorry

I can’t agree with that - I think it’s useful to see this info. For example, some GMs use an extension or a prebuilt NPC library and might not remember an NPC’s exact power level. Or, as a player, you might just want to see your numbers grow (some players like that). But ok - I’ll remove it.

> For the portrait, given the way HP work and considering support for CI, this is a little trickier. There is a significant threshold reached at 0, but there are so many alternatives to the character sheet during most of the play, I find a lot of players have them closed or are using a specialised sheet anyway (like Categorised). The Actor tokens should update to reflect status (reeling, etc.). I'm not against the character sheet reflecting it, but something that was more accurate - perhaps showing a progression in one half (say, left side) to 0, and then on the right half showing progression to -5xHP? Just an idea.

> Colours for the numbers and other things should be selectable by the user. People with colour blindness, for example, may struggle with red against certain backgrounds, and there are other conditions with other challenges, so setting a default is great but then being able to adjust is better.

I agree, we can improve it, but it’s probably better to do it later, once we’ve battle-tested it. I have a few more ideas (for example, adding a clickable icon button that represents stages from -1×HP to -5×HP, both to show the current state and to allow users to roll against HP thresholds), but I think it’s better to implement them step by step over time.

What do you think?

daynin:
@mjeffw @rinickolous @Farmeroz do you want me to remove CP info?

Farmeroz:
I'm late (well, early, it's a time zone thing :-) ), but I agree with rinickolous and StefanLeng.  Thank you, daynin for the work you're doing and for the discussion :-)

rinickolous:
I think this is good to merge, waiting for @mjeffw to confirm.

Regarding the TypeScript migration @daynin , we don't currently have a dedicated migration branch, our next major update branch is [develop/v13-only](https://github.com/crnormand/gurps/tree/develop/v13-only). We don't really have a full consensus on where and how types should be structured, but if you can take a look at the existing ones and have any suggestions those would be much appreciated.

PR #2462, opened 2026-01-03
Modern UI: Add Move Mode functionality to new sheet

rinickolous:
This PR adds the ability to open the Move Mode editor window and change the move mode to the "Modern UI" sheet.

I'm thinking of maybe abstracting out the custom dropdown element CSS (which I just copied from the existing maneuver and posture implementations) into some sort of template which the dropdowns would then use, but haven't included it in this PR yet. Let me know if you think that'd be a good idea.

### DIFF
--- a/templates/actor/modern/encumbrance-table.hbs
+++ b/templates/actor/modern/encumbrance-table.hbs
@@ -40,36 +40,36 @@
     {{/each}}
     <div class='ms-enc-mode'>
       <label for='move-mode'>{{localize 'GURPS.moveMode'}}:</label>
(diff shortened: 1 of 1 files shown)


daynin (approved):
I'm ok with the change, but we should address the indentation issue

mjeffw:
I fixed it. Copilot offered me the fix and I accepted it. You used tabs instead of spaces.

daynin:
I think we should add at least an .editorconfig file to the project to avoid issues like this. I’ll try to do it in the next PRs.

PR #2465, opened 2026-01-05
Bugfix/2463 shotgun type weapons calculate rof results incorrectly

mjeffw:


module/dierolls/compute-potential-hits.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @param optionalArgs.shots The number of shots fired.
+ * @param marginOfSuccess The margin of success from the attack roll.
+ * @param chatdata.rof The rate of fire text to display in chat.
+ * @param chatdata.rcl The recoil value of the weapon.
+ * @param chatdata.rofrcl The actual number of hits possible given the rate of fire and recoil.
+ */
+export function computePotentialHits(
+  optionalArgs: { obj: { rcl: string; rof: string }; shots?: number },
+  marginOfSuccess: number,
+  chatdata: Record<string, unknown>
```
daynin: Consider adding more specific types to:
1. Make it a bit more self-descriptive
2. Reflect actual data structure for `chatdata`

```typescript
export type Weapon = {
  rcl: string
  rof: string
}

export type RoFChatData = {
  rof: string
  rcl: string
  rofrcl: number
}
```

module/dierolls/compute-potential-hits.ts
```diff
@@ -0,0 +1,38 @@
...
+  const matchesShotgun = weapon.rof.match(SHOTGUN_ROF)
+
+  const rofMultiplier = matchesShotgun ? parseInt(matchesShotgun.groups!.projectiles) : 1
+  const currentRoF = numberShotsFired * rofMultiplier
+  const rofText = matchesShotgun ? `${numberShotsFired}x${matchesShotgun.groups!.projectiles}` : currentRoF.toString()
+
+  const potentialHits = Math.floor(marginOfSuccess / weaponRecoil) + 1
+  const actualHits = Math.min(currentRoF, potentialHits)
+
+  chatdata['rof'] = rofText
```
daynin: Consider returning data instead of mutating. It'll be easier to test and maintain. And we won't need `beforeEach` step in tests

module/dierolls/compute-potential-hits.ts
```diff
@@ -0,0 +1,38 @@
+// Regex to match shotgun rate of fire formats like '3x9', '2*5', etc.
+const SHOTGUN_ROF = /(?<rof>\d+)[×xX\*](?<projectiles>\d+)/
+
+/**
+ * Calclates potential hits based on weapon stats and margin of success, and returns data for chat message.
```
daynin:
```suggestion
 * Calculates potential hits based on weapon stats and margin of success, and returns data for chat message.
```

module/dierolls/compute-potential-hits.ts
```diff
@@ -0,0 +1,38 @@
...
+  const numberShotsFired = Math.min(optionalArgs.shots || 1, maxRateOfFire)
+
+  // Support shotgun RoF (3x9, for example).
+  const matchesShotgun = weapon.rof.match(SHOTGUN_ROF)
+
+  const rofMultiplier = matchesShotgun ? parseInt(matchesShotgun.groups!.projectiles) : 1
+  const currentRoF = numberShotsFired * rofMultiplier
+  const rofText = matchesShotgun ? `${numberShotsFired}x${matchesShotgun.groups!.projectiles}` : currentRoF.toString()
+
+  const potentialHits = Math.floor(marginOfSuccess / weaponRecoil) + 1
```
daynin: We can have a potential issue here with division by zero. Probably better to validate it before division:

```
const weaponRecoil = Math.max(1, parseInt(weapon.rcl) || 1)
```

module/dierolls/compute-potential-hits.ts
rinickolous: Good suggestion. I second this.

module/dierolls/compute-potential-hits.ts
```diff
@@ -0,0 +1,38 @@
...
+export function computePotentialHits(
+  optionalArgs: { obj: { rcl: string; rof: string }; shots?: number },
+  marginOfSuccess: number,
+  chatdata: Record<string, unknown>
+): void {
+  const weapon = optionalArgs.obj
+
+  // Could be '3x9' for shotguns, but we want just the first number.
+  const maxRateOfFire = parseInt(weapon.rof)
+  const weaponRecoil = parseInt(weapon.rcl)
```
rinickolous: Just parroting @daynin 's suggestion below in suggestion format. In practice, GURPS weapons will almost never have a Recoil of 0, but better to be safe.
```suggestion
  const weaponRecoil = Math.max(1, parseInt(weapon.rcl) || 1)
```

module/dierolls/compute-potential-hits.ts
rinickolous: I think I agree with this in general.

module/dierolls/compute-potential-hits.ts
mjeffw: Not really arguing, but I'm taking existing functionality and trying to muck with it as little as possible. I don't have the complete definition of either `chatdata` or `optionalArgs` (and they are heavily overloaded in any case). I'll make the change but it will NOT be a real description of either structure.

module/dierolls/compute-potential-hits.ts
mjeffw: Yes, I agree in general as well. Again, changing existing functionality is a little more fraught with errors, so I just chose not to modify the original code that much. But I agree that I can probably make this change without too much risk.

module/dierolls/compute-potential-hits.ts
mjeffw: Agreed.

PR #2470, opened 2026-01-07
Merge: main -> develop/v13-only

rinickolous:
Not as clean a merge as I would've liked, but I think I have most things covered. To minimize required changes, what amounts to some duplicate or close duplicate interfaces remain.

This effectively just adds the modern-ui changes from `main` to `develop/v13-only`. Lines _removed_ have been kept to a minimum.

Notes:
- Some new lang keys introduced by Modern Sheet settings are not fully localized.
- To minimize changes, I added a temporary `gurpslink2` function which I think mimics the behavior of the old `gurpslink` function. A change made in `develop/v13-only` would otherwise cause issues with rendering the Modern sheet. Should be looked into.
- The Modern sheet uses some DialogV1 prompts, which will need to be migrated.


module/utilities/gurpslink.js
```diff
@@ -36,3 +36,36 @@ export function gurpslink(str /*, clrdmods = true, returnActions = false*/) {
   output += str
   return output
 }
+
+// NOTE: temporary, just here to fix an issue with the Modern Sheet. Should be replaced.
+export function gurpslink2(str, returnActions = false) {
```
mjeffw: Do we really need this new function? Couldn't we just restore the `returnActions` flag on the original function's argument list? They appear identical to me if `returnActions` is false.

test/lib/extract-otfs.test.ts
mjeffw, line 1: Test files should be placed in a subfolder of the `test` top level folder that reflects the location of the method being tested. In this case you intend to test the `extractOtfs` handlebars helper, so the test file should go into `test/lib`.

test/extract-otfs.test.ts
```diff
@@ -0,0 +1,211 @@
...
+  'weapon-block',
+  'controlroll',
+  'roll',
+  'damage',
+  'derivedroll',
+  'deriveddamage',
+  'chat',
+]
+
+function extractOtfs(text: string) {
```
mjeffw: This is a bad unit test. We don't test functions written in the test itself; we always test the production code.  (The big problem is that if `extractOtfs` changes in `moustachewax.js` there is no way to know we also need to update this local copy of the function.)

I understand that the problem here is that `extractOtfs` is a handlebars helper, and you can't directly test that. 

The correct approach would be to refactor the handlebars helper code into a standalone top-level function (either in `moustachewax.js` or a new file) and have the Handlebars helper call that function. Then import that function here and test it.

See Handlebars helper `tracker` for an example.

test/modern-ui/entity-config.test.ts
```diff
@@ -0,0 +1,148 @@
...
+  createArgs?: () => string[]
+}
+
+interface ModifierConfiguration {
+  isReaction: boolean
+}
+
+const mockLocalize = (key: string) => key
+
+const entityConfigurations: EntityConfiguration[] = [
```
mjeffw: @daynin @rinickolous 
The code should be importing `EntityConfiguration` from `module/actor/modern/entity-config.ts' and testing that, not a local copy.

test/modern-ui/entity-config.test.ts
```diff
@@ -0,0 +1,148 @@
...
+  {
+    entityName: 'ranged',
+    path: 'system.ranged',
+    editMethod: 'editRanged',
+    localeKey: 'GURPS.ranged',
+    createArgs: () => [mockLocalize('GURPS.ranged'), '10', '1d']
+  }
+]
+
+const modifierConfigurations: ModifierConfiguration[] = [
```
mjeffw: @daynin @rinickolous 
Same comment: import `ModifierConfiguration` from `entity-config.ts` and test it.

mjeffw (changes requested):

test/extract-otfs.test.ts
daynin: you're absolutely right, my bad, I just added it to start write tests before extracting it, band forgot to extract it later... I'll fix it tomorrow. Same for comments below

test/modern-ui/entity-config.test.ts
rinickolous: Just a note for @daynin too: Our test setup file already creates a mock localization function namespaced to game.i18n, so there's no need to make one specifically in your test file.

rinickolous:
All comments by @mjeffw have been addressed.

global.d.ts
mjeffw, line 1: I'm not a Typescript expert by any means, but is it recommended that so many type declarations go into a global.d.ts file? I would expect a bare minimum and really just legacy javascript and external library types.

daynin: No, it's not recommended, eventually we'll keep there only types for external deps that doesn't have their onw type files

PR #2478, opened 2026-01-10
TS migration, quality of life improvements

daynin:
I think this is the best time to enable ESLint, Prettier, and TypeScript checking. It's safer to do now than after the migration is complete, and it'll help us migrate easier by catching issues early.

  Also migrated the modern UI to Foundry 13 API and added more types, getting rid of all the type errors we had.

  What's included:
  - ESLint + Prettier + TypeScript strict checking enabled
  - Pre-push hook runs all checks automatically (npm run verify)
  - Modern sheet now uses Foundry 13 ApplicationV2 API

### DIFF
--- a/module/actor/modern/sheet.ts
+++ b/module/actor/modern/sheet.ts
@@ -276,14 +276,10 @@ export class GurpsActorModernSheet extends SheetBase {
   static async #onImportActor(this: GurpsActorModernSheet, event: PointerEvent): Promise<void> {
     event.preventDefault()
 
-    switch (this.actor.type) {
-      case 'character':
-        return new ActorImporter(this.actor).importActor()
-      case 'enemy':
-      case 'characterV2':
-        return GURPS.modules.Importer.importGCS(this.actor)
-      default:
-        throw new Error(`Invalid actor type for import: ${this.actor.type}`)
+    if (this.actor.isOfType('characterV2')) {
+      await GURPS.modules.Importer.importerPrompt(this.actor)
+    } else {
+      return new ActorImporter(this.actor).importActor()
     }
   }
 
--- a/global.d.ts
+++ b/global.d.ts
@@ -6,6 +6,7 @@ import { ActorV1Model } from 'module/actor/legacy/actorv1-interface.ts'
 import { GurpsCombatant } from 'module/combat/combatant.ts'
 import DamageChat from 'module/damage/damagechat.js'
 import { GurpsActiveEffect } from 'module/effects/active-effect.js'
+import { Importer } from 'module/importer/index.ts'
 import { EquipmentModel } from 'module/item/data/equipment.ts'
 import { SkillModel } from 'module/item/data/skill.ts'
 import { SpellModel } from 'module/item/data/spell.ts'
@@ -30,9 +31,7 @@ declare global {
     }
 
     modules: {
-      Importer: {
-        importGCS(actor: Actor.Implementation): Promise<void>
-      }
+      Importer: typeof Importer
       Pdf: {
         handleOnPdf(event: Event): void
       }


module/types/guards.ts
rinickolous, line 1: The functions themselves are fine, but I'm not sure about the location. Putting them under `module/types/` may imply this is somewhere to put type files.

module/actor/modern/sheet.ts
```diff
@@ -22,88 +32,125 @@ export function countItems(record: Record<string, EntityComponentBase> | undefin
...
+      editMoveMode: GurpsActorModernSheet.#onEditMoveMode,
+    },
+  }
+
+  static override PARTS: Record<string, HandlebarsTemplatePart> = {
+    sheet: {
+      template: 'systems/gurps/templates/actor/actor-modern-sheet.hbs',
+      scrollable: ['.ms-body .tab'],
+    },
   }
```
rinickolous, line 95: This implementation of PARTS is fine for now if you just want to minimize changes, but I would want to separate the distinct parts of the sheet at a later point.

module/actor/modern/sheet.ts
```diff
@@ -22,88 +32,125 @@ export function countItems(record: Record<string, EntityComponentBase> | undefin
...
 
-  // @ts-expect-error - Template returns modern sheet path which extends base templates
-  override get template() {
-    if (!game.user!.isGM && this.actor.limited) return 'systems/gurps/templates/actor/actor-sheet-gcs-limited.hbs'
+  get template(): string {
+    if (!game.user!.isGM && this.actor.limited) {
+      return 'systems/gurps/templates/actor/actor-sheet-gcs-limited.hbs'
```
rinickolous, line 99: Good note for later: A restricted/view-only version of the modern sheet may be something we want later.

module/actor/modern/sheet.ts
```diff
@@ -133,49 +181,155 @@ export class GurpsActorModernSheet extends GurpsActorSheet {
...
+
+    // Bind dropdown handlers
+    this.#bindPostureActions(html)
+    this.#bindManeuverActions(html)
+    this.#bindMoveModeActions(html)
+
+    // Bind entity CRUD actions
+    this.#bindEntityCrudActions(html)
+
+    // Wire up OTF rollable elements (requires jQuery wrapper for now)
```
rinickolous: Add a TODO: note saying to remove the jQuery wrapper when the appropriate conversion is done.
(it's easier, at least for me, to track TODO notes if they are formatted as `TODO: ...`)

module/actor/modern/sheet.ts
```diff
@@ -133,49 +181,155 @@ export class GurpsActorModernSheet extends GurpsActorSheet {
...
+      })
+    })
+
+    // Wire rollable elements
+    html.querySelectorAll<HTMLElement>('.rollable').forEach(element => {
+      element.addEventListener('click', this.#onClickRoll.bind(this))
+    })
+  }
+
+  // ============================================
```
rinickolous: Just a minor thing, but I've been sticking to one separator and section heading comment style in my code (`/* -*40 */`) and would prefer consistency in that.

module/actor/modern/sheet.ts
```diff
@@ -133,49 +181,155 @@ export class GurpsActorModernSheet extends GurpsActorSheet {
...
 
-  bindPostureActions(html: JQuery): void {
+  static async #onImportActor(this: GurpsActorModernSheet, event: PointerEvent): Promise<void> {
+    event.preventDefault()
+    switch (this.actor.type) {
+      case 'character':
+        return new ActorImporter(this.actor).importActor()
+      case 'enemy':
+      case 'characterV2':
+        // @ts-expect-error GURPS.modules is dynamically populated
```
rinickolous: Good point here. We should define the type of `GURPS.modules`

module/actor/modern/sheet.ts
```diff
@@ -243,4 +376,243 @@ export class GurpsActorModernSheet extends GurpsActorSheet {
...
+            obj.situation = html.find('.situation').val()
+            await actor.internalUpdate({ [path]: obj })
+          },
+        },
+      },
+      default: 'one',
+    }).render(true)
+  }
+
+  async editItem(
```
rinickolous, line 625: This function should not be used at all in v13 as far as I'm concerned. We should be doing away with non-Foundry-Item Items entirely at this point.

templates/actor/modern/header/ms-portrait.hbs
```diff
@@ -1,6 +1,6 @@
 <div class='ms-portrait-wrapper'>
   <div class='ms-portrait'>
-    <img src='{{actor.img}}' data-edit='img' title='Total: {{system.totalpoints.total}}&#10;Unspent: {{system.totalpoints.unspent}}&#10;&#10;Attributes: {{system.totalpoints.attributes}}&#10;Advantages: {{system.totalpoints.ads}}&#10;Disadvantages: {{system.totalpoints.disads}}&#10;Quirks: {{system.totalpoints.quirks}}&#10;Skills: {{system.totalpoints.skills}}&#10;Spells: {{system.totalpoints.spells}}&#10;Race: {{system.totalpoints.race}}' />
+    <img src='{{actor.img}}' data-action='configurePrototypeToken' title='Total: {{system.totalpoints.total}}&#10;Unspent: {{system.totalpoints.unspent}}&#10;&#10;Attributes: {{system.totalpoints.attributes}}&#10;Advantages: {{system.totalpoints.ads}}&#10;Disadvantages: {{system.totalpoints.disads}}&#10;Quirks: {{system.totalpoints.quirks}}&#10;Skills: {{system.totalpoints.skills}}&#10;Spells: {{system.totalpoints.spells}}&#10;Race: {{system.totalpoints.race}}' />
```
rinickolous: this should be `editImage`, not `configurePrototypeToken`

module/actor/data/base.ts
```diff
@@ -1,5 +1,5 @@
-import TypeDataModel = foundry.abstract.TypeDataModel
-import fields = foundry.data.fields
+const TypeDataModel = foundry.abstract.TypeDataModel
+const fields = foundry.data.fields
```
rinickolous: Why is this preferred to import?

module/damage/damagecalculator.js
```diff
@@ -264,10 +258,6 @@ export class CompositeDamageCalculator {
     this._armorDivisor = value
   }
 
-  get useArmorDivisor() {
```
rinickolous, line 245: Just looks like this function is being removed. Why?

module/types/foundry/index.ts
rinickolous, line 1: What's the reasoning behind this?

module/actor/modern/sheet.ts
mjeffw: I use a plugin to VSCode that looks for all comments that start with "TODO" and builds an action item list for me, so I agree with this.

module/actor/modern/sheet.ts
daynin: For now I wanted to minimize changes for sure. I didn't want to refactor it to much along with other changes, just wanted to make it work with V2 api

module/actor/modern/sheet.ts
daynin: I''ll remove jQuery in the following PRs, no need in adding TODOs in code

module/actor/modern/sheet.ts
daynin: Ok, I don't mind

module/actor/modern/sheet.ts
daynin: Yes, sure, but again I din't want to produce a lot of new types/changes here

module/types/foundry/index.ts
daynin: We don't use it in code

module/actor/data/base.ts
daynin: Sorry, didn't get the question. I see that's the comment is outdated, maybe you pointed to something else

templates/actor/modern/header/ms-portrait.hbs
daynin: done

module/actor/modern/sheet.ts
daynin: I see. You're right. But let's refactor it later, in the following PRs

module/types/guards.ts
daynin: Moved it to `module/utilities`

module/actor/data/base.ts
rinickolous: The question is "why is declaring `const TypeDataModel = foundry.abstract.TypeDataModel` preferable to `import TypeDataModel = foundry.abstract.TypeDataModel`, though I see you switched to declaring these types in one place and then importing them like we do everything else afterwards.

module/actor/modern/sheet.ts
rinickolous: Yes, that's fine, I wasn't suggesting that you do this as part of this PR.

module/types/foundry/index.ts
rinickolous: I think we may be getting two comments confused here. "What's the reasoning behind this?" refers to the foundry native types being declared in one place and then imported everywhere from that location. Is this some sort of future-proofing in case the namespace changes again or something?

daynin: Got it, yes, we can't rely on fvtt-types, they are not complete and they lie. We still can use them wrapping them in our type wrappers. I put them there to easily add ad-hoc wwrappers exposing same modules. It'll allow us reduce amount of changes in the future

module/actor/data/base.ts
daynin: I see, it's incorrect, we should use our types here. I think I might lost the change during rebase. Updated. Thanks for pointing to the issue!

`import X = namespace.Y` is TypeScript's proper syntax for namespace aliasing. We should stick to it

package.json
```diff
@@ -48,8 +48,8 @@
     "format:check": "prettier --check .",
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
-    "tdd": "NODE_OPTIONS=\"--experimental-vm-modules\" jest --coverage --config=jest.config.mjs",
-    "test": "NODE_OPTIONS=\"--experimental-vm-modules\" jest --config=jest.config.mjs",
+    "tdd": "NODE_OPTIONS=\"--experimental-vm-modules --disable-warning=ExperimentalWarning\" jest --coverage --config=jest.config.mjs",
+    "test": "NODE_OPTIONS=\"--experimental-vm-modules --disable-warning=ExperimentalWarning\" jest --config=jest.config.mjs",
```
rinickolous, line 52: glad to see this

mjeffw:
As mentioned I want the bugs fixed in 0.18.11 done first. Then we'll get to this.

mjeffw:
A general comment:

- I would prefer splitting large automatic formatting and linting out of actual code changes. The huge number of changes may make it hard to review actual code changes.

package.json
mjeffw, line 1: I'd prefer new tooling to be discussed and agreed-upon before implementation. Keep in mind that @crnormand and I don't know all of these fancy JS/TS/etc tools and want some understanding of what is being introduced and what to expect once implemented.

Maybe include a comment here or in the wiki or Discord, but I'm mostly guessing about what `husky` does; `eslint` seems more familiar but was there any discussion about what rules to enforce, etc.?

I've been relying on VSC's automatic reformatting using `prettier` rather than something built into the project.

I see the `prepare` script just invokes `husky` -- does that just run the git pre-commit hooks?

When would I need to run the `verify` script?

You know, questions like that... :)

package.json
```diff
@@ -41,13 +44,19 @@
...
+    "format": "prettier --write .",
+    "format:check": "prettier --check .",
+    "lint": "eslint .",
+    "lint:fix": "eslint . --fix",
+    "tdd": "NODE_OPTIONS=\"--experimental-vm-modules --disable-warning=ExperimentalWarning\" jest --coverage --config=jest.config.mjs",
+    "test": "NODE_OPTIONS=\"--experimental-vm-modules --disable-warning=ExperimentalWarning\" jest --config=jest.config.mjs",
+    "typecheck": "tsc --noEmit",
+    "verify": "run-s format:check lint typecheck test",
     "wach:styles": "sass --watch --update --style=expanded gurps.scss:dist/gurps.css",
     "watch": "run-p wach:styles watch:code watch:static",
```
mjeffw: Is there a typo on this line? `wach` instead of `watch`?

module/actor/data/base.ts
mjeffw: I (and @crnormand) might need some pointers or examples of this once we get into this code.

.editorconfig
mjeffw, line 1: I assume this applies to VS Code? And if you are using another IDE, we can rely on the prettier rules?

eslint.config.js
```diff
@@ -0,0 +1,225 @@
...
+  setProperty: 'readonly',
+  getProperty: 'readonly',
+  hasProperty: 'readonly',
+  expandObject: 'readonly',
+  flattenObject: 'readonly',
+  randomID: 'readonly',
+  debounce: 'readonly',
+
+  // GURPS system
+  GURPS: 'readonly',
```
mjeffw, line 63: What does `readonly` mean in this context? I mean, we are adding all kinds of functions and data to the global GURPS object.

package.json
daynin: ESLint is a linter - it verifies that the code follows certain rules (it includes more rules than Prettier). It can also find dead code and prevent keeping it unless it’s explicitly marked as such. Basically, it helps keep the code cleaner and ensures it stays consistent.

Prettier is a formatter. It just formats code according to its rules.

TS type checker - I made it stricter to prevent adding new `any` types and to encourage covering the code with proper types.

Overall, ESLint + TypeScript checks are the gold standard for TypeScript projects.

Husky runs `verify` as a pre-push hook to prevent pushing unverified code to branches.

> I would prefer splitting large automatic formatting and linting out of actual code changes

I agree - I’ll do it in new PRs. I added it here to make sure my changes are already verified.

> I've been relying on VSC's automatic reformatting using prettier rather than something built into the project.

VS Code will automatically follow configs (eslint, prittier and typescript). You'll see same errors/warnings that you can get from running `npm run verify`

> I see the prepare script just invokes husky -- does that just run the git pre-commit hooks?

`prepare` just initialize husky for the project, actual hook is `pre-push`

> When would I need to run the verify script?

You don't need to run this script manually. It will be run automatically during pushes

package.json
daynin: Fixed

.editorconfig
daynin: Yes, most of editors will use eslint/prettier/ts configs automatically, but even if they won't, we have `pre-push` that runs it

eslint.config.js
daynin: It means that these global variables are readonly and eslint won't allow use modifying them. Without this we'll have two issues:
- eslint doesn't allow using unconfigured global vars
- we'll be able to mistakenly rewrite these vars

package.json
daynin: > was there any discussion about what rules to enforce, etc.

No, sorry, there wasn’t any discussion. I just included rules that fit our current coding style. I think we can adjust them over time.

I mainly wanted to create a foundation to keep the code style consistent and improve linting gradually. I just thought it’s better to have at least something we can rely on, especially since we’re going to do a huge refactor

.editorconfig
rinickolous: `.editorconfig` isn't VS Code-specific. I'm using Neovim and it has built in support. Most modern IDEs do as far as I know.

daynin: Yes, that's right, I use nvim as well, it works out of the box

PR #2480, opened 2026-01-10
Feature: GCA Importer for new Actor Data Model

rinickolous:
`dev-utilities/gca5.xsd` is a file taken from GCA5 which acts as a reference for the GCA5 internal data model.

The changes to `importer/migrate.ts` fix an issue where only the local (loaded) copy of the old settings is removed rather than the version in the db.

The rest is the GCA importer.


utils/gca5.xsd
daynin, line 1: We already have `utils` directory, I think we just should put this file there instead of creating one more dir with similar name. It may confuse people seeing both dirs and not knowing where to put their files

module/importer/gca-importer/importer.ts
```diff
@@ -0,0 +1,768 @@
...
+    this.#importProfile()
+    this.#importHitLocations()
+    this.#importItems()
+    this.#importPointTotals()
+    this.#importMiscValues()
+
+    if (actor) {
+      // When importing into existing actor, save count and uses for equipment with ignoreImportQty flag
+      const savedEquipmentCounts = this.#saveEquipmentCountsIfNecessary(
+        actor.items.contents.filter(item => item.type === 'equipmentV2') as Item.OfType<'equipmentV2'>[]
```
daynin: I'd suggest avoiding type casting. In typescript it may lead to miscasting in some cases. Let's try to use proper types at the first place or use type guards to verify if something has a certain type

module/importer/gca-importer/importer.ts
```diff
@@ -0,0 +1,768 @@
...
+      )
+
+      // When importing into existing actor, delete only imported items
+      await this.#deleteImportedItems(actor)
+
+      // Update actor with new system data and create new items
```
daynin, line 86: [note] Just a style concern. Do we really need such comments? I mean it's clear from the code and comments can quickly become stale. I think we should write comments for non-obvious things and try to write self descriptive code for the rest

module/importer/gca-importer/importer.ts
```diff
@@ -0,0 +1,768 @@
...
+      actor = (await Actor.create({
+        _id,
+        name,
+        type,
+        img: this.img,
+        system: this.output as any,
+        items: this.items as any,
+      })) as Actor.OfType<'characterV2'> | undefined
+    }
+    if (!actor) {
```
daynin, line 108: I'd suggest to avoid implicit type casting. Let's use type guards instead like `!isDefined(actor)`. It's cleaner and will help in preventing runtime errors. Same for places like that down below

module/importer/gca-importer/importer.ts
```diff
@@ -0,0 +1,768 @@
...
+    }
+
+    this.img = `data:image/png;base64,${this.input.vitals?.portraitimage}.png`
+  }
+
+  /* ---------------------------------------- */
+
+  #importAttributes() {
+    this.output.attributes = { ST: {}, DX: {}, IQ: {}, HT: {} }
+    for (let key of ['ST', 'DX', 'IQ', 'HT', 'QN', 'WILL', 'PERCEPTION'] as const) {
```
daynin, line 199: [note] we can put this array to actual const outside of the method to avoid recreation of it. Not a big deal, it'll be jit-optimized, but anyway, it's simple change, we could do this

module/importer/gca-importer/importer.ts
```diff
@@ -0,0 +1,768 @@
...
+      const attribute = this.input.traits.attributes.find(attr => attr.name.toLowerCase() === key.toLowerCase())
+      if (attribute) {
```
daynin: [note] we could simplify it a bit:

```ts
        this.output.attributes[key === 'PERCEPTION' ? 'PER' : key] = {
          import: attribute?.score ?? 10,
          points: attribute?.points ?? 0,
        }
```

or

```ts
        this.output.attributes[key === 'PERCEPTION' ? 'PER' : key] = isDefined(attribute) ? {
          import: attribute?.score ?? 10,
          points: attribute?.points ?? 0,
        } : { import: 10, points: 0 }
```

Depending of what exact behavior we want here

Same for similar places below

module/importer/gca-importer/importer.ts
```diff
@@ -0,0 +1,768 @@
...
+
+  /* ---------------------------------------- */
+
+  #importHitLocations() {
+    this.output.additionalresources ||= {}
+    this.output.additionalresources.bodyplan = this.input.bodytype ?? 'Humanoid'
+
+    const table = HitLocations.hitlocationDictionary?.[this.input.bodytype ?? 'Humanoid']
+    this.output.hitlocationsV2 = []
+    this.input.hitlocationtable.hitlocationlines?.forEach(location => {
```
daynin: It would be cleaner to rewrite it to map like `this.output.hitlocationsV2 = this.output.hitlocationsV2.concat(location.map...`

module/importer/gca-importer/importer.ts
```diff
@@ -0,0 +1,768 @@
...
+          if (!weapon._id || typeof weapon._id !== 'string') {
+            console.error('GURPS | Failed to import weapon: No _id set.')
+            console.error(weapon)
+            return acc
+          }
+          acc[weapon._id] = weapon
+          return acc
+        },
+        {}
+      ) as any
```
daynin: let's try to avoid `any` in new typescript code

module/importer/gca-importer/importer.ts
```diff
@@ -0,0 +1,768 @@
...
+      _id,
+      type,
+      name,
+      system: {
+        ...system,
+        fea: component,
+      },
+    }
+
+    this.items.push(item)
```
daynin, line 581: I think we need to choose one:
1. We return value and push it to `this.items` outside of the method (that's more preferable to me)
2. We do not return value and push it to `this.items` inside of the method

utils/gca5.xsd
rinickolous: This isn't a new directory, `dev-utilities` has existed for some time. However, I agree that having 2 directories like this is confusing. I would _prefer_ to move all the actual system code into a top-level `src/` directory. I'll suggest that in a separate PR.

module/importer/gca-importer/importer.ts
rinickolous: the `location.map` call won't work here because some members of `location` will not return valid values (i.e. if bodyLocation has no valid matching value).

utils/gca5.xsd
mjeffw: So to be clear, `utils` are scripts that are called by the GGA runtime, while `dev-utilities` are dev time utilities, typically called by the developer from the command line.

module/importer/gca-importer/importer.ts
rinickolous: cleaned up slightly anyway.

module/importer/gca-importer/importer.ts
rinickolous: There's some sort of limitation on the type system when calling `Actor.create`, set this to expect error for now.

module/importer/gca-importer/importer.ts
rinickolous: Yeah, seems like that's just vestigial. Fixed.

module/importer/gca-importer/importer.ts
rinickolous: Simplified. Thanks!

module/importer/gca-importer/importer.ts
rinickolous: Fixed, thanks.

module/importer/gca-importer/importer.ts
rinickolous: Where would I put it in that case? A constant outside the class? A property of the class?

utils/gca5.xsd
rinickolous: @mjeffw this file is neither, it's just a reference file. Where should it go?

module/importer/gca-importer/importer.ts
daynin: I'd say, a constant outside of the class. In this case it'll be truly static and won't be  recreated in between calls

daynin (approved):

module/importer/gca-importer/importer.ts
```diff
@@ -0,0 +1,769 @@
...
+    if (!actor) {
+      throw new Error('Failed to create GURPS actor during import.')
+    }
+
+    return actor
+  }
+
+  /* ---------------------------------------- */
+
+  #restoreEquipmentCountsAndUses(savedEquipmentCounts: Map<string, { quantity: number; uses: number }>) {
```
mjeffw: There are similar functions in the GCS importer, could they be shared?

module/importer/gca-importer/importer.ts
```diff
@@ -0,0 +1,769 @@
...
+      if (eqt && eqt.importFrom === 'GCA' && eqt.ignoreImportQty && eqt.importid) {
+        savedEquipmentCounts.set(eqt.importid, { quantity: eqt.count, uses: eqt.uses })
+      }
+    })
+    return savedEquipmentCounts
+  }
+
+  /* ---------------------------------------- */
+
+  #importPortrait() {
```
mjeffw: This could be refactored into a shared function with the GCS importer.

module/importer/gca-importer/importer.ts
mjeffw: In general I think there is a lot of code here that could be refactored and shared across GCS and GCA importers.

PR #2483, opened 2026-01-11
fix: update quotedAttackName function to handle quotes and add tests

### LINKED ISSUE #2461: Dashes in Ranged Weapon names used as Modifiers
If you have a ranged weapon with a - and number in the name, it's treated as a modifier. Not sure if this is intentional.

I encountered this by having a gun named AK-98. Clicking to roll the skill level caused it to have a -98 modifier on the roll (It took me awhile to realize why).

Image

Image

mjeffw:
Fixes #2461

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,7 +4,12 @@
 
 ### Bugfixes
 
-- Dashes in Ranged Weapon names used as Modifiers #2461
+- Modern UI sheet uses `GURPS.LastActor` as actor target instead of sheet actor \#2472
+- Modern UI sheet cuts off tall portraits \#2474
+- Can't change encumbrance manually on modern UI sheet \#2475
+- Modern UI sheet does not display resource trackers \#2476
+- Weapon attack tagged modifiers are sometimes added to damage rolls \#2482
+- Dashes in Ranged Weapon names used as Modifiers \#2461
 
 ## Release 0.18.11 01/09/2026
 
@@ -15,7 +20,7 @@
 ### Bugfixes
 
 - Missing lang tags \#2450
-- Can't roll from Actor Sheet when Canvas disabled #2449
+- Can't roll from Actor Sheet when Canvas disabled \#2449
 - ModToolTip with transparent BG \#2448
 - Cannot create Active Effects from Actor Sheet \#2447
 - Fright Check dialog text unreadable (v18.10) \#2446
@@ -161,7 +166,7 @@
 - Improved targeting of damage #2178
 - Auto-Apply of modifiers should require Roll Confirmation to be active #2151
 
-### Bugfixes:
+### Bugfixes
 
 - Apply Damage fails when any token in the scene has no actor #2169
 - Some skills not recognized if not quoted in OtF #2163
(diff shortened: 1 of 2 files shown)


lib/utilities.js
```diff
@@ -574,13 +574,17 @@ export function arrayBuffertoBase64(buffer) {
...
+ */
 export function quotedAttackName(item) {
-  let q = '"'
-  let n = item.name
-  if (n.includes(q)) q = "'"
-  if (!!item.mode) n = n + ' (' + item.mode + ')'
-  if (n.includes(' ')) n = q + n + q
-  return n
+  const quoteCharacter = item.name.includes(DOUBLE_QUOTE) ? SINGLE_QUOTE : DOUBLE_QUOTE
+  let n = !!item.mode ? item.name + ' (' + item.mode + ')' : item.name
```
daynin: I'd suggest avoiding short names like `n`

lib/utilities.js
```diff
@@ -574,13 +574,17 @@ export function arrayBuffertoBase64(buffer) {
...
+ * @returns {string}
+ */
 export function quotedAttackName(item) {
-  let q = '"'
-  let n = item.name
-  if (n.includes(q)) q = "'"
-  if (!!item.mode) n = n + ' (' + item.mode + ')'
-  if (n.includes(' ')) n = q + n + q
-  return n
+  const quoteCharacter = item.name.includes(DOUBLE_QUOTE) ? SINGLE_QUOTE : DOUBLE_QUOTE
```
daynin: Probably it's still can fail if a skill contains both quote types - `Someone's "Super Power"`? I'd suggest using "`" character or even square brackets. I think we should reconsider some syntax to avoid issues like that

lib/utilities.js
```diff
@@ -574,13 +574,17 @@ export function arrayBuffertoBase64(buffer) {
...
+  let n = !!item.mode ? item.name + ' (' + item.mode + ')' : item.name
+  return quoteCharacter + n + quoteCharacter
```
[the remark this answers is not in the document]
daynin: Just noticed that copilot suggested the same about both types of quotes

test/utilities.test.ts
```diff
@@ -284,6 +284,16 @@ describe('utilities', () => {
       const item = { name: 'Attack', mode: 'Mode' }
       expect(utilities.quotedAttackName(item)).toBe('"Attack (Mode)"')
     })
+
+    test('should return quoted attack name without mode', () => {
```
daynin, line 288: I think we should add a test that reproduced exact bug from the issue.I'll be easier to find relation between the change and the issue later

lib/utilities.js
mjeffw: We need to overhaul OTFs in the near future. I don't want to go to another delimiter until we do, as the change will ripple throughout the code.

test/utilities.test.ts
mjeffw: done

mjeffw:
> @mjeffw is this ready for re-review?

Yes

rinickolous:
@mjeffw do make sure to add the bugfix to the change log before merging

PR #2490, opened 2026-01-13
Bugfix: setEventFlags command shows user an error for some chat commands if the event is not writeable

rinickolous:
This is in reference to [a bug report for [Token Action HUD GURPS](https://github.com/rinickolous/fvtt-token-action-hud-gurps/issues/7). I was able to recreate the error messages the user complained about, which resulted from the Event attached to the command being a PointerEvent instead of the expected event type, and the `shiftKey` and `ctrlKey` properties of that event not being writeable.

This PR preserves the error message in the log but reduces it to a warning and hides it from the UI. As far as I can tell, the message turns into a whisper as expected and the `mergeObject` call may be redundant, but am reluctant to remove it in case it breaks something else.

I expect that this (sort of hacky) fix will be overwritten following any refactor of our chat processing (which I think I'll get to basically once #2478 is merged), so this is a temporary fix.

### DIFF
--- a/module/chat.js
+++ b/module/chat.js
@@ -305,9 +305,10 @@ class ChatProcessorRegistry {
     this.msgs.oldQuiet = quiet
     this.msgs.data.whisper = [game.user?.id]
     // HACK: This try / catch prevents errors from showing to the user in cases where properties shiftKey and/or ctrlKey of `this.msgs.event`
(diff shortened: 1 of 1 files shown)


mjeffw:
> I expect that this (sort of hacky) fix will be overwritten following any refactor of our chat processing (which I think I'll get to basically once #2478 is merged), so this is a temporary fix.

Actually, I thought that @crnormand might take the rewrite of the chat processing, since he wrote most of it and needs a friendly reintroduction to the code.

PR #2497, opened 2026-01-15
Fix DR parsing to use named groups and handle unspecified locations i…

mjeffw:
…n GurpsActor

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -14,6 +14,7 @@
 - Fixed drag-and-drop on the combat tracker. \#2486
 - Fixed drag-and-drop on the Modifier Bucket. \#2486
 - Movement range color indicators do not appear on gridless maps \#2489
+- DR Modifiers not working using Foundry Items \#2496
 
 ## Release 0.18.11 01/09/2026
 


module/utilities/contextmenu.js
```diff
@@ -1,7 +1,5 @@
-// Context Menu
-// COMPATIBILITY: v12
-// export default class GgaContextMenu extends foundry.applications.ux.ContextMenu {
-export default class GgaContextMenu extends ContextMenu {
+
+export default class GgaContextMenu extends foundry.applications.ux.ContextMenu {
```
rinickolous, line 2: I was going to say that this change would break FoundryVTT v12 compatibility, but just checked and confirmed that our current `main` branch is _already_ incompatible with FoundryVTT v12. Commit cf9a9613 changed the GurpsActorSheet superclass namespace from global to `foundry.appv1.sheets.ActorSheet` which produces an error in v12. In fact, it looks like `0.18.11` (the current `release` branch) already has issues with the declaration of `GurpsRuler`. The declaration of the class is being read even if the `if` condition fails. This produces an error but doesn't prevent further initialization and thus may have gone unnoticed for our v12 users.

In any case, we can't change this namespace for a v12 compatible release.

rinickolous (changes requested):
Change itself is fine but the other unrelated change pertaining to ContextMenu should be reversed.

module/utilities/contextmenu.js
mjeffw: Ah, I have been changing the namespaces here and there, I forgot that was v12 incompatible.

mjeffw: I've opened another issue to deal with the v12 compatibility.

PR #2515, opened 2026-01-18
Bugfix/restore v12 functionality

mjeffw:


system.json
rinickolous, line 1: What's up with the reshuffle here?

mjeffw: Accident. I'll fix it.

PR #2526, opened 2026-01-22
TS migration, quality of life improvements

rinickolous:
This is just a copy of #2478 but is on this repo instead of a fork so we can make changes more easily. Normal description follows:

I think this is the best time to enable ESLint, Prettier, and TypeScript checking. It's safer to do now than after the migration is complete, and it'll help us migrate easier by catching issues early.

Also migrated the modern UI to Foundry 13 API and added more types, getting rid of all the type errors we had.

What's included:

- ESLint + Prettier + TypeScript strict checking enabled
- Pre-push hook runs all checks automatically (npm run verify)
- Modern sheet now uses Foundry 13 ApplicationV2 API


daynin:
Leave it to me. I'll do it

rinickolous:
We've already merged another PR which aims to replicate the changes within this one, give or take some. Closing for now.

PR #2531, opened 2026-01-26
Fix modifiers not added on item drag-and-drop

666f78:
Fix bug where dragging items did not add bonuses, itemModifiers, modifierTags, or addToQuickRoll.
They are now added correctly after drag-and-drop.

### DIFF
--- a/module/actor/actor.js
+++ b/module/actor/actor.js
@@ -1534,6 +1534,11 @@ export class GurpsActor extends Actor {
         actorComp.duringotf = data.system[data.itemSysKey].duringotf
         actorComp.passotf = data.system[data.itemSysKey].passotf
         actorComp.failotf = data.system[data.itemSysKey].failotf
+        actorComp.bonuses = data.system.bonuses || ''
+        actorComp.itemModifiers = data.system.itemModifiers || ''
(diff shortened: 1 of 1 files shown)


rinickolous (commented):
I'm happy to merge this if it solves an issue you've been experiencing, but I'd just like to understand what the expected / actual behavior is a little better. Could you give me some steps to follow to see the issue at play so I can compare?

666f78:
> I'm happy to merge this if it solves an issue you've been experiencing, but I'd just like to understand what the expected / actual behavior is a little better. Could you give me some steps to follow to see the issue at play so I can compare?

Sure — here are the steps.

1. In Settings, enable Foundry Items.
2. Create an Item and add a modifier to it.
3. Drag & drop the item onto a Character Sheet.
4. Open the item on the character sheet.

Actual: The modifier is not displayed.
Expected: The modifier should be displayed.

666f78:
> Give me an example of a modifier that you've tried.


https://github.com/user-attachments/assets/ef6b02c0-a2f5-4442-8758-a12757f5369c

daynin:
LGTM. But if it's possible let's add tests that cover this change

PR #2532, opened 2026-01-26
Minor improvements to Modern Sheet

666f78:
- Added a recalc button to lifting and moving.

image

- Made the Acc modifier clickable.

image

image

### DIFF
--- a/module/actor/modern/sheet.ts
+++ b/module/actor/modern/sheet.ts
@@ -164,6 +164,7 @@ export class GurpsActorModernSheet extends GurpsActorSheet {
     this.bindManeuverActions(html)
     this.bindEncumbranceActions(html)
     this.bindEffectActions(html)
+    this.bindRangedAccActions(html)
     this.bindLiftingActions(html)
     this.bindEntityCrudActions(html)
   }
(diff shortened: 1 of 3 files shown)


666f78:
> The Acc modifier should be styled like other modifiers on the sheet. Take a look at the "Modifiers" table under the Basic tab for an example.
> 
> I did add automatic calculation of Swing, Thrust, and Basic Lift that can be toggled on or off via a new setting: "Auto Update Strength". If on, the button becomes redundant, but a GM may decide to turn that off and request manual updates, that your function provides.
> 
> I do, however, want to unify the code that does the calculation. That code will be in the `prepareDerivedData` method of the Actor. I'd like you to call the same code when I merge it into `main`.
> 
> That method is `GurpsActor.updateStrengthBasedAttributes`.

I want to point out that contributors opening PRs may not have full context on all existing or upcoming features.
Instead of rhetorical questions, it would be more helpful to explicitly say when something is already implemented or planned. That saves time and avoids unnecessary friction.

Since this functionality already exists in main, I don’t see much value in continuing to iterate on this PR, so I’ll close it.
The remaining changes are relatively minor (such as adding or adjusting a button) and should be straightforward to implement directly alongside the existing code.

PR #2533, opened 2026-01-28
Feature/strength calculators

mjeffw:
I've got to add this to the GCA import as well.

This Feature automatically updates Basic Lift, Thrust, and Swing when ST is edited. It also imports the basic "sw+x" type damage for muscle-powered weapons.


rinickolous (approved):
No issues as far as I can see.

PR #2540, opened 2026-02-02
Bugfix: various bugs after main -> develop/v13-only merge

rinickolous:
- Fixed errors when merging global.d.ts causing `fvtt-types` configuration to break
- Removed duplicate/outdated handlebars helpers
- Removed deprecated import settings references
- Removed deprecated function calls
- Fixed inconsistent types returned for Maneuver function calls for different Actor types
- Fixed Modern sheet not rendering when Maneuver is set
- Replaced references to `Settings.SYSTEM_NAME` with `GURPS.SYSTEM_NAME`

### DIFF
--- a/module/actor/gurps-actor.ts
+++ b/module/actor/gurps-actor.ts
@@ -3409,7 +3409,7 @@ class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> impleme
    * @deprecated Actor v1 only.
    */
   private _getCurrentMoveMode() {
-    let move = this.modelV1.move as Record<string, MoveMode>
+    let move = (this.modelV1?.move ?? {}) as Record<string, MoveMode>
     let current = Object.values(move).find(it => it.default)
     if (!current && Object.keys(move).length > 0) return move['00000']
     return current


module/actor/gurps-actor.ts
```diff
@@ -82,6 +84,8 @@ interface EquipmentDropData {
 }
 
 class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> implements ActorV1Interface {
+  strengthCalculator: StrengthCalculator | null = null
```
mjeffw: The whole dynamic Strength calculation is being removed from 0.18.12; please go ahead and remove it here as well.

module/actor/gurps-actor.ts
```diff
@@ -533,6 +537,74 @@ class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> impleme
     }
   }
 
+  /* ---------------------------------------- */
+
+  updateStrengthBasedAttributes() {
```
mjeffw: Remove.

module/actor/gurps-actor.ts
```diff
@@ -533,6 +537,74 @@ class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> impleme
...
+    this.system.swing = this.strengthCalculator.calculateSwingDamage()
+  }
+
+  /* ---------------------------------------- */
+
+  /**
+   * Update the basic lift and recalculate encumbrance levels and lifting.
+   * @param {*} basicLift
+   */
+  _setBasicLift(basicLift: unknown): void {
```
mjeffw: Remove

module/actor/gurps-actor.ts
```diff
@@ -533,6 +537,74 @@ class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> impleme
...
+    this.system.liftingmoving.carryonback = basicLift * 15
+    // @ts-expect-error: old character types in use
+    this.system.liftingmoving.runningshove = basicLift * 24
+    // @ts-expect-error: old character types in use
+    this.system.liftingmoving.shiftslightly = basicLift * 50
+  }
+
+  /* ---------------------------------------- */
+
+  async updateAndPersistStrengthBasedAttributes() {
```
mjeffw: Remove

module/token-actions.js
```diff
@@ -383,15 +383,17 @@ export class TokenActions {
    */
   async addModifiers() {
     const addModifier = mod => {
-      if (!maneuverModifiers.includes(mod) && !allModifiers.includes(mod)) {
+      if (!maneuverModifiers.includes(mod) && !allModifiers.has(mod)) {
```
mjeffw: Educate me -- what is the difference between Array.includes and Array.has?

rinickolous: `.has` is equivalent to `.includes`, but `.has` applies to Sets while `.includes` applies to Arrays. `system.conditions.usermods` is of type Array for Character V1 but of type Set for Character V2.

This change was supposed to go alongside normalizing `allModifiers` to always be a set, but this may not be as appropriate as setting it to an Array instead.

A subsequent commit normalizes `allModifiers` to an Array instead.

PR #2541, opened 2026-02-03
TS migration, quality of life improvements (take 2)

rinickolous:
This is a limited version of #2526 and does not include some changes such as elimination of type assertions. However, if @daynin can let me know how to highlight those for changing so I can keep track of them easily then I can add that to this PR too.

List of changes:
- Added eslint and .editorconfig files for managing code standards.
- Fixed all newly presenting errors and warnings resulting from the above.
- Replaced direct Foundry API imports with internal ones which we can then change as needed.
- Migrated Modern Sheet and Modern NPC sheet to ApplicationV2

### DIFF
--- a/module/actor/actor-importer.js
+++ b/module/actor/actor-importer.js
@@ -4,6 +4,7 @@ import * as HitLocations from '../hitlocation/hitlocation.js'
 import { ImportSettings } from '../importer/index.js'
 import { SmartImporter } from '../smart-importer.js'
 import { calculateEncumbranceLevels, readXmlText } from '../utilities/import-utilities.js'
+
 import {
   Advantage,
   Encumbrance,
@@ -630,6 +631,7 @@ export class ActorImporter {
     }
 
     let lm = {}
+
     lm.basiclift = readXmlText(json.basiclift)
     lm.carryonback = readXmlText(json.carryonback)
     lm.onehandedlift = readXmlText(json.onehandedlift)
@@ -719,6 +721,7 @@ export class ActorImporter {
   importTraitsfromGCA(json) {
     if (!json) return
     let ts = {}
+
     ts.race = readXmlText(json.race)
     ts.height = readXmlText(json.height)
     ts.weight = readXmlText(json.weight)
@@ -734,6 +737,7 @@ export class ActorImporter {
     ts.techlevel = readXmlText(json.tl)
     // <appearance type="string">@GENDER, Eyes: @EYES, Hair: @HAIR, Skin: @SKIN</appearance>
     let a = readXmlText(json.appearance)
+
     ts.appearance = a
 
     try {
@@ -798,11 +802,13 @@ export class ActorImporter {
    */
   async importBaseAdvantagesFromGCA(datalist, json) {
     if (!json) return
+
     for (let key in json) {
       if (key.startsWith('id-')) {
         // Allows us to skip over junk elements created by xml->json code, and only select the skills.
         let j = json[key]
         let a = new Advantage()
+
         a.name = readXmlText(j.name)
         a.originalName = readXmlText(j.name)
         a.points = this.intFrom(j.points)
@@ -811,6 +817,7 @@ export class ActorImporter {
         a.uuid = readXmlText(j.uuid)
         a.parentuuid = readXmlText(j.parentuuid)
         let old = this._findElementIn('ads', a.uuid)
+
         this._migrateOtfsAndNotes(old, a, readXmlText(j.vtt_notes))
         a = await this._processItemFrom(a, 'GCA')
         datalist.push(a)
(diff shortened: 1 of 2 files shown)


module/actor/data/character.ts
```diff
@@ -634,17 +650,20 @@ class CharacterModel extends BaseActorModel<CharacterSchema> {
 
   #prepareUserModifiers() {
     this.parent.items.forEach(item => {
-      if (!(item as GurpsItemV2).isOfType('featureV2', 'skillV2', 'spellV2', 'equipmentV2')) return
+      if (!(item as Item.Implementation).isOfType('featureV2', 'skillV2', 'spellV2', 'equipmentV2')) return
```
mjeffw, line 653: Question (Help me learn something new):

Why is `item as Item.Implementation` preferable to `item as GurpsItemV2` when we know our Item implementation is via the `GurpsItemV2` class?

module/actor/actor-importer.js
```diff
@@ -128,19 +138,22 @@ export class ActorImporter {
...
       r = JSON.parse(json)
       this.json = r
-    } catch (err) {
+    } catch {
       msg.push(game.i18n.localize('GURPS.importNoJSONDetected'))
       exit = true
     }
-    if (!!r) {
+
+    if (r) {
```
mjeffw, line 145: I'd love it if there was a linting rule that disallowed single character variable names.

module/damage/damagechat.js
```diff
@@ -80,21 +84,26 @@ export default class DamageChat {
     const actor = game.actors.get(dropData.actorid)
 
     switch (dropData.type) {
-      case DragDropType.DAMAGE:
+      case 'damageItem':
```
mjeffw: Shouldn't this be `DragDropType.DAMAGE`?

module/damage/damagechat.js
```diff
@@ -411,9 +448,9 @@ export default class DamageChat {
     }
 
     messageData['flags.gurps.transfer'] = {
-      type: DragDropType.DAMAGE,
+      type: 'damageItem',
```
mjeffw: `DragDropType.DAMAGE`?

test/lib/parselink.test.ts
```diff
@@ -1847,43 +1856,6 @@ describe('parseLink', () => {
         expect.stringContaining("data-otf='A:Throwing Axe -2 stunned'><b>A:</b>Throwing</span> Axe -2 stunned")
       )
     })
-
```
mjeffw, line 1850: Why was this code excluded? I think these tests are still valid.

module/chat/tracker.js
```diff
@@ -21,72 +23,87 @@ export default class TrackerChatProcessor extends ChatProcessor {
...
 
-    if (!!m[6]) {
+    if (m[6]) {
       // reset -- Damage Tracker's reset to zero
-      let value = !!theTracker.isDamageTracker ? parseInt(theTracker.min) : parseInt(theTracker.max)
+      let value = theTracker.isDamageTracker ? parseInt(theTracker.min) : parseInt(theTracker.max)
+
       //if (!!theTracker.isDamageTracker) max = 0
-      await actor.update({ ['data.additionalresources.tracker.' + theTrackerKey + '.value']: value })
+      await actor.update({ ['system.additionalresources.tracker.' + theTrackerKey + '.value']: value })
```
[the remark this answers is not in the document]
mjeffw: It would be nice to make this consistent.

module/actor/data/character.ts
rinickolous: `Item.Implementation` is a reference pointing to where `GurpsItemV2` is defined as our `Item` documentClass in `global.d.ts`. If we ever change the name of the class or change which class we use, we won't have to go through every mention of `GurpsItemV2` and change that. This also means we don't need to import `GurpsItemV2` into this file.

module/damage/damagechat.js
rinickolous: Yes, must've been an overwritten correction. Fixed.

module/chat/tracker.js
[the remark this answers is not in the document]
rinickolous: All consistent now as far as I can tell

module/actor/actor-importer.js
rinickolous: There is one, actually. I've automatically identified ~600 places where single character variables names are used. Would you like me to include corrections for that in this PR or a subsequent one?

test/lib/parselink.test.ts
rinickolous: Accidental removal. Re-added.

module/actor/actor-importer.js
mjeffw: Subsequent one.

PR #2549, opened 2026-02-09
Refactor: Implement domain code demarcation

rinickolous:
- Implemented TS paths across the code-base to remove extensive `../` imports
- Moved all code to be compiled by TSC into `src/`
- Moved all static files used within Foundry to `static/`
- Split reusable utilities files into `src/module/util` and `src/util/`. The former should contain only FoundryVTT-dependent utilities, while the latter should include only non-FoundryVTT-dependent utilities.
- Created `src/rules/` folder, which should include only non-FoundryVTT-dependent implementations of GURPS rules. This is near-empty now, but appropriate files should be migrated to this location over time.
- Modified build scripts to reflect the above changes
- Removed some unused dependencies
- Moved `fvtt-types/configuration` module to dedicated file (per fvtt-types dev recommendation)
- Converted some utility files from JS to TS

Most of the files marked as "changed" in this PR are simply moved to a different location (marked as "renamed").

Some files are moved and have had their `import` lines changed to match the new TS paths config, or to reflect their new location.

Some files previously did not fall under linting/formatting scope but now do, and therefore contain ESLint / Prettier automatic changes. For some ESLint rules (namely the minimum variable name rule), there are also manual corrections implemented.

Some files are not recognized by git as being moved and are marked as deleted and created elsewhere. This is especially the case for a few files which were converted from JS to TS as part of this PR.

### DIFF
--- a/static/templates/apply-damage/select-token.hbs
+++ b/static/templates/apply-damage/select-token.hbs
@@ -11,7 +11,7 @@
     {{#each tokens}}
       <label>
         <input type='checkbox' name='tokens' value='{{this.id}}' class='exclusive-checkbox' />
-        {{this.name}}
+        {{this.actor.name}}
       </label>
     {{/each}}
   </div>


dev/scripts/find-missing-keys.js
mjeffw, line 1: Obsolete; delete this file.

dev/scripts/i18n-todo.js
mjeffw, line 1: Obsolete; delete this file.

dev/scripts/reverse-lang.js
mjeffw, line 1: Obsolete; delete this file.

src/module/combat-tracker/index.ts
```diff
@@ -1,7 +1,7 @@
-import { GurpsModule } from '../gurps-module.ts'
+import type { GurpsModule } from '@gurps-types/gurps-module.js'
 
 import { addManeuverListeners } from './maneuver-menu.js'
-import { renderCombatTracker } from './render-combat-tracker.ts'
+import { renderCombatTracker } from './render-combat-tracker.js'
```
mjeffw, line 4: It's a shame you can't import the actual file, only the transpiled one. Is there a way to get this back?

src/rules/compute-potential-hits.ts
mjeffw, line 1: Candidate file for the `rules` directory

src/module/damage/damagecalculator.js
mjeffw, line 1: Candidate for moving to the `rules` directory. It probably requires a little bit of work to disconnect it from Foundry code.

src/module/util/select-target.ts
```diff
@@ -0,0 +1,69 @@
+/**
+ * Multiple tokens are selected -- prompt the user for which tokens to apply some function to.
+ */
+export default async function selectTarget(
```
mjeffw, line 4: This appears to be translated from JS to TS. I'd expect these functions to be tested before the PR is created.

src/rules/gcs/dice.ts
mjeffw, line 1: This can go into the `rules` folder *now*, with no other work.

src/util/ranges.js
mjeffw, line 1: This is another file that might be able to be moved to `rules`.

src/rules/size-speed-range-table.js
mjeffw, line 1: Might be able to move to `rules`.

src/rules/compute-potential-hits.ts
mjeffw: For any of these comments, I don't expect you to move them to `rules` if they require any modifications. The main thing is that they can't import any Foundry entities like Actor, Token, etc.

src/module/damage/damagecalculator.js
mjeffw: For any of these comments, I don't expect you to move them to rules if they require any modifications. The main thing is that they can't import any Foundry entities like Actor, Token, etc.

mjeffw (changes requested):

src/module/combat-tracker/index.ts
rinickolous: I don't think this is actually a matter of being unable to import the actual file. It's just that non-relative imports with `.ts` extensions don't get their paths automatically changed to `.js` during compilation.
However, this path is relative so we should be able to change it back. Probably just got caught in Find & Replace

src/util/ranges.js
rinickolous: It references `game.settings`, so I won't move it for this PR, but I agree it should be moved after being updated to separate it from that reference.

rinickolous: Added a TODO in the file.

src/module/damage/damagecalculator.js
rinickolous: Currently references Foundry code by proxy (`token-actions.js` -> `miscellaneous-settings.js`) and `./settings.js` -> `gurps-settings-application.js`) so not moving yet, but adding TODO

src/module/util/select-target.ts
rinickolous: There are no code changes in the file, just type decorations. Nevertheless, tested and found an unrelated issue (Foundry v13 Tokens seem to have a `this.name` value equal to `this.actor.type` for some reason, so corrected this.

PR #2550, opened 2026-02-09
Fix: AoA (Determined Ranged) does not appear in Bucket and Modifiers Pop-up (1.0.0)

### LINKED ISSUE #2529: Feature Correction: All out Attack (Determined) Ranged +1 does not appear in the Modifier Bucket or Manoeuvre options (only the Melee +4 version does)
As per subject, need to add All out Attack (Determined) Ranged +1 to both the Modifier Bucket options and the Manoeuvre options.

Image

Image

Image

Image

rinickolous:
Also changed how the Modifier Bucket gets attack modifiers. The related functions previously returned long string literals with a bunch of newlines which were later broken up into arrays. This may have been desirable for easier editing or what-not at some point but I think it's messy so I changed those functions to just return arrays.

This is the 1.0.0 version of this fix, the 0.18 version is in a separate PR.

Closes #2529

### DIFF
--- a/src/module/token/token-hud.ts
+++ b/src/module/token/token-hud.ts
@@ -36,7 +36,13 @@ export class GurpsTokenHUDV2 extends foundry.applications.hud.TokenHUD {
       // TODO: revise any to specific type
       maneuvers: Object.entries(GURPS.Maneuvers.getAll()).map(([id, maneuver]: [string, any]) => {
         return {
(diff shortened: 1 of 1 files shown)


rinickolous:
Noticed an issue just now, don't merge yet.

daynin (approved):

PR #2552, opened 2026-02-09
GCS Actor and Item Data Models

rinickolous:
Many features in this PR are not ready for production and are hidden by default. To enable UI creation of the new Actor and Item types, run the following command in the FoundryVTT console after the world loads:

```javascript
game.settings.set("gurps", "developerMode", true)
```

This PR should not modify any existing functionality unrelated to the GCS objects described below. Any new errors or bugs are likely already present in `develop/v13-only`. Although the functionality described here is incomplete, the code should be safe to merge and remain dormant in release versions until it is ready.

## `GcsCharacter` Data Model

This is currently mostly a stub, containing only persistent data and minimal processing via a few accessors. Even these accessors (namely `attributes` and `attributeDefinitions`) could arguably be cached via `prepareBaseData` or a similar method, which would significantly improve performance given how frequently they are expected to be queried.

## `GcsLoot` Data Model

This is a stub representing the “Loot” sheet added to GCS some time ago. It is an actor that can contain Items and has minimal descriptive data with no other substantive functionality.

Including this at this stage highlights a future migration issue:

Assuming that all methods related to “Character”-type Actors (whether legacy `character`, `characterV2`, or `gcsCharacter`) apply uniformly to all actors will not be sustainable. Actor-type-specific methods and data should reside in their respective DataModel classes wherever possible. While this is not feasible for the legacy `character` type, the migration described above should align with its deprecation and removal.

To that end, any interfaces (Sheets, etc.) interacting with GCS Actor and Item types should interface with the system object as exclusively as possible.

This PR includes several new type guards to preserve legacy functionality. At present, these are primarily sufficient to satisfy ESLint and TSC rather than to provide functional parity between legacy and GCS Actor/Item types.

## Item Data Models

These include `GcsTraitModel`, `GcsTraitModifierModel`, `GcsSkillModel`, `GcsSpellModel`, `GcsEquipmentModel`, `GcsEquipmentModifierModel`, and `GcsNoteModel`.

Like `GcsCharacter`, these are mostly stubs consisting of declared persistent data schemas and stub interface implementations. Each class defines a `metadata` static property identical to that of `BaseItemModel`. Their superclass, `GcsBaseItemModel`, closely mirrors `BaseItemModel` and provides separation from legacy implementations.

Some DataModel classes include accessors marked as placeholders. In most cases, these should be replaced with derived but cached properties resolved during a child method of `prepareData`. Performance considerations (such as replacing accessors with cached properties evaluated during data preparation) should remain a priority in further development.

## GCS `Attribute`, `AttributeDefinition`, and `AttributeDefinition` DataModels.

These are mostly stubs with persistent-data schemas as above. `AttributeDefinition`, however, includes calls to the `ScriptResolver` class (see below).

## Scripting Module

GCS uses a stripped-down internal JS parser and interpreter extensively to compute derived values in rich-text fields and base attribute values. Earlier versions of GCS used a minimal custom scripting language limited to simple expressions.

Because GGA runs inside a JS application, sandboxing presents additional challenges: users can access the console and execute arbitrary code, potentially affecting application state and complicating GCS feature parity.

Previously explored options included:

- Using `eval`: Risky, difficult to debug, effectively impossible to sandbox correctly, and inadequate for achieving GCS parity.
- Using the `Function` constructor: Slightly safer than `eval` due to positional parameters, but still insufficient in terms of sandboxing and parity.
- Combining the above with Foundry’s `AsyncWorker` helper: Improved isolation via JS Workers, but incompatible with GCS feature requirements. GCS adapters rely on dynamically changing properties and objects that reference their source instances. JS Workers cannot accept hydrated objects, class instances, or functions, making this approach impractical.

This module includes `ScriptResolver` and `ScriptInterpreter` classes, as well as adapter classes representing GCS entities (Characters, Attributes, and Items—though Items are not yet implemented), and limited globally accessible functions such as `Math` and `Dice`.

`ScriptInterpreter` uses the `acorn` library to parse JS into AST objects, which are interpreted and executed in a tightly sandboxed environment. Scripts accept an initial context, effectively treated as `globalThis` during execution. By restricting this context, we can support a substantial subset of JS functionality expected by users and already supported in GCS.

While `ScriptInterpreter` can modify application state if given DOM or game data objects, its use within `ScriptResolver` is **not** intended to do so. Users expecting GCS scripts to behave like FoundryVTT macros may initially find this distinction unclear.

`ScriptResolver` constructs the execution environment for `ScriptInterpreter` and resolves its output to a specified type. For example, `AttributeDefinition` uses it to compute initial attribute values.

## `GcsMeleeWeapon` and `GcsRangedWeapon`

These are currently mostly stubs. They are PseudoDocuments that users will be able to embed within Items.

## Features

The following classes represent item features that can be attached to GCS Items and are currently mostly stubs:

`AttributeBonus`, `ConditionalModifier`, `ContainedWeightReduction`, `CostReduction`, `DRBonus`, `ReactionBonus`, `SkillBonus`, `SkillPointBonus`, `SpellBonus`, `SpellPointBonus`, `TraitBonus`, and `WeaponBonus`.

Each is intended to behave according to its name and mirror GCS functionality once implemented.

## Prerequisites

The following classes represent prerequisites that users can attach to GCS Items and are currently mostly stubs (most now include an `isSatisfied` accessor central to their functionality):

`AttributePrereq`, `ContainedQuantityPrereq`, `ContainedWeightPrereq`, `EquippedEquipmentPrereq`, `PrereqList`, `ScriptPrereq`, `SkillPrereq`, `SpellPrereq`, and `TraitPrereq`.

They are intended to display errors on the character sheet when unsatisfied, informing users that their character configuration is invalid.

## Study

These items mark progress in learning a Skill or upgrading a Trait. They appear to serve only a record-keeping function in GCS and are currently stubs.

## Weight

Similar to the `Length` class, this DataModel represents a specified weight, including value and units. Static and instance methods mirror those of `Length`, supporting unit-agnostic addition, unit conversion, conversion to primitive types (`string`, `number`), and parsing from `string` and `number`.

Both `Weight` and `Length` are accompanied by custom field types `WeightField` and `LengthField`. These extend `EmbeddedDataField` and automatically parse `string` and `number` input into `Weight` and `Length` instances.

## Criteria Checkers

These classes evaluate an input value (`string`, `string[]`, `number`, or `Weight`) against defined criteria. They are used by Features and Prerequisites.

## A Note on `ArrayField`

In this PR, I have avoided using `ArrayField` for arrays of non-primitive types (`ArrayField<StringField>` is acceptable and is used in a few places). Instead — following guidance from experienced FoundryVTT developers — I have used `TypedObjectField` with randomly generated keys via `foundry.utils.randomID`.

For fields backed by `TypedObjectField`, ordering should be handled through a `sort` property (or similar) on the contained item.

In practice, `ArrayField` is difficult to work with because its contents cannot be partially updated; modifying a single element requires replacing the entire array. Avoiding it at this stage — and committing to a consistent object-backed pseudo-array pattern — should make future development and migrations significantly easier.

### DIFF
--- a/src/module/config/attributes.ts
+++ b/src/module/config/attributes.ts
@@ -42,7 +42,6 @@ const defaultAttributes = (): Record<string, AttrData> => {
       fullName: 'Dexterity',
       costPerPoint: 20,
       costAdjustmentPerSizeMod: 0,
-      _thresholds: null,
     }),
     withId({
       sort: 2,
@@ -54,7 +53,6 @@ const defaultAttributes = (): Record<string, AttrData> => {
       fullName: 'Intelligence',
       costPerPoint: 20,
       costAdjustmentPerSizeMod: 0,
-      _thresholds: null,
     }),
     withId({
       sort: 3,
@@ -66,7 +64,6 @@ const defaultAttributes = (): Record<string, AttrData> => {
       fullName: 'Health',
       costPerPoint: 10,
       costAdjustmentPerSizeMod: 0,
-      _thresholds: null,
     }),
     withId({
       sort: 4,
@@ -78,7 +75,6 @@ const defaultAttributes = (): Record<string, AttrData> => {
       fullName: '',
       costPerPoint: 5,
       costAdjustmentPerSizeMod: 0,
-      _thresholds: null,
     }),
     withId({
       sort: 5,
@@ -90,7 +86,6 @@ const defaultAttributes = (): Record<string, AttrData> => {
       fullName: '',
       costPerPoint: 2,
       costAdjustmentPerSizeMod: 0,
-      _thresholds: null,
     }),
     withId({
       sort: 6,
@@ -102,7 +97,6 @@ const defaultAttributes = (): Record<string, AttrData> => {
       fullName: 'Perception',
       costPerPoint: 5,
       costAdjustmentPerSizeMod: 0,
-      _thresholds: null,
     }),
     withId({
       sort: 7,
@@ -114,7 +108,6 @@ const defaultAttributes = (): Record<string, AttrData> => {
       fullName: '',
       costPerPoint: 2,
       costAdjustmentPerSizeMod: 0,
-      _thresholds: null,
     }),
     withId({
       sort: 8,
@@ -126,7 +119,6 @@ const defaultAttributes = (): Record<string, AttrData> => {
       fullName: '',
       costPerPoint: 2,
       costAdjustmentPerSizeMod: 0,
-      _thresholds: null,
     }),
     withId({
       sort: 9,
@@ -138,7 +130,6 @@ const defaultAttributes = (): Record<string, AttrData> => {
       fullName: '',
       costPerPoint: 2,
       costAdjustmentPerSizeMod: 0,
-      _thresholds: null,
     }),
     withId({
       sort: 10,
@@ -150,7 +141,6 @@ const defaultAttributes = (): Record<string, AttrData> => {
       fullName: '',
       costPerPoint: 2,
       costAdjustmentPerSizeMod: 0,
-      _thresholds: null,
     }),
     withId({
       sort: 11,
@@ -162,7 +152,6 @@ const defaultAttributes = (): Record<string, AttrData> => {
       fullName: '',
       costPerPoint: 20,
       costAdjustmentPerSizeMod: 0,
-      _thresholds: null,
     }),
     withId({
       sort: 12,
@@ -174,7 +163,6 @@ const defaultAttributes = (): Record<string, AttrData> => {
       fullName: '',
       costPerPoint: 5,
       costAdjustmentPerSizeMod: 0,
-      _thresholds: null,
     }),
     withId({
       sort: 13,
--- a/src/module/config/body.ts
+++ b/src/module/config/body.ts
@@ -174,6 +174,7 @@ const defaultBodyType = (): fields.SchemaField.CreateData<GcsBody.Schema> => {
   ]
 
   return {
+    _id: foundry.utils.randomID(),
     name: 'Humanoid',
     roll: '3d',
     _locations: Object.fromEntries(


src/module/data/fields/map-field.ts
```diff
@@ -0,0 +1,384 @@
...
+
+    value = foundry.utils.applySpecialKeys(value)
+
+    if (
+      value === current ||
+      [...(value as Map<AssignmentKeyType, AssignmentValueType>)]?.equals([
+        ...(current as Map<AssignmentKeyType, AssignmentValueType>),
+      ])
+    )
+      return
```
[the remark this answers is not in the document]
rinickolous: Foundry provides the `Array#equals` function so this is fine.

mjeffw:
My instincts are to reject this PR. It violates a number of precepts I thought we had agreed to. 

- It is too large to be reviewed and understood (at least by me!) in the short amount of time that a PR should remain open.
- It introduces yet-another data model.
- It may not be implemented in a way that we could incrementally release pieces of functionality to the users.
- No one else has had any input into its implementation. (Maybe the most important reason!)

I'm trying to pull @crnormand in to get his view on it.

rinickolous:
> It is too large to be reviewed and understood (at least by me!) in the short amount of time that a PR should remain open.

Sure, it is a large PR. That is partly because the feature itself is broad, and partly because it is still at an early stage of development. The schema is not yet settled. Abstractions may be introduced and later removed. Field names and types may change as requirements become clearer.

Constraining the work to a single DataModel or narrowly defined sub-feature does not avoid that volatility. As the remaining pieces are implemented, earlier decisions are likely to need revisiting once the broader context is in place. That leads to the same models being modified again in later PRs as understanding of the feature evolves.

The components also interlock. Attributes depend on the ScriptResolver. The ScriptResolver depends on resolution adapters. The adapters depend on the classes they are adapting. Decisions in one area shape the others, which makes early isolation of individual pieces fragile while the design is still in flux.

An alternative approach seems to be introducing one DataModel at a time and releasing them incrementally. Two issues arise:

1. What practical value does a single, isolated DataModel provide?
   A GCS Equipment Item without compatible Actors is of limited use.
   A GCS Actor without Items is similarly constrained.
   A Script Resolver that is not invoked anywhere does not meaningfully serve users.

2. Incremental release increases migration churn and developer overhead.
   With a schema that is still evolving, partial releases would require follow-up migrations as abstractions shift or fields are renamed or retyped. Developers would also need to maintain transitional states and compatibility logic across versions, rather than converging on a stable design.

Migrations are sometimes necessary. Adding extra ones while the design is still fluid creates additional work for both users and developers.

A possible response is that Items and Actors could be introduced separately with an adapter layer for backwards compatibility. That approach carries its own costs:

1. An adapter layer allows partially implemented models to exist in an environment they were not designed for, increasing surface area for bugs and requiring additional maintenance.

2. Because the components interlock, those adapters and partial implementations are likely to need revision once the full system is in place, adding further developer effort.

Releasing the feature piecewise under these conditions spreads ongoing design changes across multiple PRs and migrations, rather than containing them while the requirements are still settling.

> It introduces yet-another data model.

We have already agreed that migrating toward a GCS-based DataModel is the future direction for GGA, and this has been communicated to users. This PR serves to implement that direction, so why would this be an issue?

> It may not be implemented in a way that we could incrementally release pieces of functionality to the users.

Given the interdependent nature of the system and the current fluidity of the schema, incremental release is difficult without repeated revisions to earlier components as the design matures. If there is a concrete proposal for doing this while avoiding that churn, I am open to examining it. It would need to address both the structural dependencies and the evolving requirements.

> No one else has had any input into its implementation. (Maybe the most important reason!)

This PR is the opportunity for that input. I welcome critique, requested changes, structural objections, or alternative designs.

The branch exists in this repository rather than a fork. If you would like to author changes directly, you are free to submit a PR into this branch.

src/module/action/gcs-melee-attack.ts
mjeffw, line 1: Why commit empty files?

src/module/action/gcs-ranged-attack.ts
mjeffw, line 1: Why commit empty files?

src/module/action/index.ts
```diff
@@ -1,31 +1,43 @@
...
       // TODO: localize
       label: 'Ranged Attack',
-      // @ts-expect-error: Type shows as invalid but is valid?
       documentClass: RangedAttackModel,
     },
   }
 }
 
-export const Action: GurpsModule = {
+export const ActionModule: GurpsModule = {
```
mjeffw, line 43: I prefer modules to not have a boilerplate 'Module` suffix, unless there is a compelling reason. Just like we don't add "Class" to the end of class names.

src/module/action/melee-attack.ts
```diff
@@ -5,19 +5,18 @@ import { AnyObject } from 'fvtt-types/utils'
...
-  declare mel: MeleeAttackComponent
   static override defineSchema(): MeleeAttackSchema {
     return Object.assign(super.defineSchema(), meleeAttackSchema())
   }
 
   /* ---------------------------------------- */
 
   static override get TYPE(): string {
-    return 'meleeAttack'
+    return ActionType.MeleeAttack
```
mjeffw, line 16: Thank you for using declared constants for things like this instead of hardcoded strings.

src/module/actor/data/gcs-character/attribute-threshold.ts
```diff
@@ -0,0 +1,55 @@
...
+const attributeThresholdSchema = () => {
+  return {
+    ...pseudoDocumentSchema(),
+    sort: new fields.IntegerSortField({ required: true, nullable: false, initial: 0 }),
+    state: new fields.StringField({ required: true, nullable: false }),
+    value: new fields.StringField({ required: true, nullable: false }),
+    explanation: new fields.StringField({ required: true, nullable: false }),
+    // NOTE: STUB. This field is used to store operation names (as strings) which correspond to
+    // halving values like ST, Move, etc. at the given threshold. There may be a better way of
+    // storing this information.
```
mjeffw, line 37: I've not had any time to look at this in detail, but what if they were objects with methods? The methods would implement the calculation. Maybe make a generic Operation class with a "perform" method with instances that do the right thing when perform is called.

mjeffw: This could be used extensively such as when reducing movement and other values.

src/module/actor/data/gcs-character/attribute-definition.ts
```diff
@@ -0,0 +1,190 @@
...
+      initial: GcsAttributePlacement.Automatic,
+    }),
+    name: new fields.StringField({ required: true, nullable: false }),
+    fullName: new fields.StringField({ required: true, nullable: false }),
+    // NOTE: This is parsed as JS code, but no type yet exists for this.
+    // TODO: Create dedicated JS code type.
+    base: new fields.JavaScriptField({ required: true, nullable: false }),
+    costPerPoint: new fields.NumberField({ required: true, nullable: false, initial: 0 }),
+    // NOTE: Should be displayed as a percentage
+    costAdjPerSm: new fields.NumberField({ required: true, nullable: false, initial: 0 }),
```
mjeffw: I prefer being explicit on variable naming. I have no idea what `costAdjPerSm` means, and I should have an idea of what it is just on how it is named. (It looks like "Cost adjustment per <something>" but what is "Sm"?)

src/module/actor/data/gcs-character/attribute-definition.ts
```diff
@@ -0,0 +1,190 @@
...
+    )
+  }
+
+  get allowsDecimal(): boolean {
+    return this.type === AttributeType.Decimal || this.type === AttributeType.DecimalRef
+  }
+
+  /* ---------------------------------------- */
+
+  get kind(): GcsAttributeKind {
```
mjeffw, line 127: Why wouldn't the attribute definition have a `kind` field? It would be simpler to just return `this.kind`.

src/module/actor/data/gcs-character/attribute-definition.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  baseValue(att: GcsAttribute): number {
+    if (this.isSeparator) return 0
+
+    return ScriptResolver.resolveToNumber(att.actor, ScriptAttribute.newProvider(att), this.base)
+  }
+
+  /* ---------------------------------------- */
+
+  get isPrimary(): boolean {
```
mjeffw, line 68: There's a lot of code like `if (this.type === Thing) return value`. Consider the `type` field itself storing whether it is a Pool or Primary or Secondary or Separator, ...

That would simplify the code to do `return type.isPrimary` or 'return type.isPool`, etc...

src/module/actor/data/gcs-character/sheet-settings.ts
```diff
@@ -0,0 +1,127 @@
+import { fields } from '@gurps-types/foundry/index.js'
+import { defaultAttributes } from '@module/config/attributes.js'
+import { defaultBodyType } from '@module/config/body.js'
+import { WeightUnit, LengthUnit } from '@module/data/common/index.js'
+
+import { GcsAttributeDefinition } from './attribute-definition.js'
+import { GcsBody } from './body.js'
+
+enum DamageProgression {
```
mjeffw, line 10: When I did that failed Damage progression thing in 0.18.x, I made each DamageProgression strategy its own object and put the calculation of swing, thrust, and lift as methods of that object. I prefer that to the GCS approach of defining an enum and then having a switch statement elsewhere in the codebase do the calculation.

mjeffw: This would even allow for a user-defined strength progression.

src/module/actor/test-actor-sheet.ts
mjeffw, line 1: What's the purpose of the TestActor sheet? What does it do that an existing sheet doesn't?

src/module/config/attributes.ts
```diff
@@ -0,0 +1,389 @@
+import { DataModel } from '@gurps-types/foundry/index.js'
+import { GcsAttributeDefinition } from '@module/actor/data/gcs-character/attribute-definition.js'
+
+// NOTE: These IDs are entirely arbitrary and I'm not convinced they need
+// to be defined. They are defined here solely so that we don't have to run
+// foundry.util.randomID() several times when a character is initialized, but
+// this config will be replaced with a system setting which will override them
+// anyway.
```
mjeffw: First: why worry about running `foundry.util.randomID()`? 

Second: if we're going to use artificial IDs anyway, is there any reason not to make them strings that are memorable? For example, in unit tests, I often give IDs that name the field or attribute. So, `st: 'strength', dx: 'dexterity'` and so on.

src/module/data/criteria/string-criteria.ts
```diff
@@ -0,0 +1,192 @@
...
+  Any = 'any',
+  Is = 'is',
+  IsNot = 'isNot',
+  Contains = 'contains',
+  DoesNotContain = 'doesNotContain',
+  StartsWith = 'startsWith',
+  DoesNotStartWith = 'doesNotStartWith',
+  EndsWith = 'endsWith',
+  DoesNotEndWith = 'doesNotEndWith',
+}
```
mjeffw, line 14: For this and NumberCriteria, consider using object values instead of enums and switch statements.

src/module/data/criteria/string-criteria.ts
```diff
@@ -0,0 +1,192 @@
...
+
+const stringCriteriaSchema = () => {
+  return {
+    compare: new fields.StringField({
+      required: true,
+      nullable: false,
+      choices: Object.values(StringComparison),
+      initial: StringComparison.Any,
+    }),
+    qualifier: new fields.StringField({ required: true, nullable: true }),
```
mjeffw, line 101: I'm not happy with this variable name but I don't have a suggestion. "qualifier" doesn't tell me anything. 

The object has a comparison and  a string, and contains methods that execute the comparison between an argument and the qualifier... "value" or "base" or "leftValue" (versus "rightValue" for the argument) night be better.

src/module/data/criteria/number-criteria.ts
mjeffw, line 1: How are NumberCriteria and StringCriteria used in the system?

src/module/features/contained-weight-reduction.ts
```diff
@@ -0,0 +1,31 @@
...
+    return FeatureType.ContainedWeightReduction
+  }
+}
+
+/* ---------------------------------------- */
+
+const containedWeightReductionSchema = () => {
+  return {
+    // NOTE: Should render as percentage
+    reduction: new fields.NumberField({ required: true, nullable: false }),
```
mjeffw, line 23: Does Foundry allow you to specify your own validate logic? It seems like we could use that to get a number from (0-1] (assuming that is what is needed for this type).

src/module/actor/data/gcs-character/attribute-definition.ts
rinickolous: This is just a case of copying the name from GCS. It's cost adjustment per Size Modifier. I can change it to be clearer.

src/module/actor/test-actor-sheet.ts
rinickolous: I'm just using it to test stuff out during development, it's not meant for production. It could be moved somewhere more appropriate like the Dev module I suppose

src/module/config/attributes.ts
rinickolous: I did consider this, but I think I ran into issues with Collection complaining if its keys are not valid 16 character IDs

src/module/data/criteria/string-criteria.ts
rinickolous: I'm not sure I understand

src/module/data/criteria/number-criteria.ts
rinickolous: They're used in Features and Prereqs to facilitate discriminating based on object attributes. 
For example, TraitPrereq can be specified to complain if it doesn't see a Trait with name `foo`, level at least `2`, and whose notes do not include `bar`

src/module/features/contained-weight-reduction.ts
rinickolous: There's nothing about the data that needs limiting (actually there is, this value should be 0-80 but that's a separate issue), it's just how it renders that needs to be specified so that the field visually communicates that it represents a percentage.

src/module/actor/data/gcs-character/attribute-definition.ts
rinickolous: That would work if there wasn't so much that determines the attribute's position. We have
- Straightforwardly checking the type. Works for separators and pools. 
- Checking manual position override (if not automatic)
- If automatic, verifying whether the attribute base value script is just a number. 

I can document this more clearly because it's bit jumbled, but I don't think hiding the required complexity elsewhere is the right move.

src/module/actor/data/gcs-character/attribute-threshold.ts
rinickolous: That's an interesting approach. I'll need to think about this a bit more

src/module/action/index.ts
rinickolous: I agree. The exception here is because the name `Action`is used in the same file for the type `Action<Type extends ActionType> =... `, and the other TypedPseudoDocuments use the same naming scheme in their respective index files e.g. `Prereq<Type extends... `

src/module/action/gcs-ranged-attack.ts
rinickolous: Just didn't get around to implementing this before deciding to submit, but put them there to come back to later.

src/module/actor/data/gcs-character/attribute-definition.ts
rinickolous: For the same reason the accessors this accessor calls exist. The value is derived from other factors and is not to be treated as persistent.

src/module/config/attributes.ts
mjeffw: Ah, okay. Well, it could be `strength-------` and `dexterity-------` and so on. But it doesn't really matter, it just makes it easier if during testing/debugging one wants to refer to an attribute by key.

The first question still stands -- if we're going to have random strings as IDs even while testing, why not call `random`? Why worry about performance during testing (and besides it is probably too fast to matter).

src/module/features/contained-weight-reduction.ts
mjeffw: Oh, so the value is a percentage such that 20 => 20%. I was thinking 20% would be 0.2.

src/module/data/criteria/string-criteria.ts
mjeffw: Instead of something like this:
```ts
    switch (this.compare) {
      case StringComparison.Any:
        return true
      case StringComparison.Is:
        return lowerValue === lowerQualifier
...
```

You'd do something like this:
```ts
  return this.compare(value)
```

Because `StringComparison` entries are actual objects with the comparison behavior built in. Adding new comparisons would be just defining the new Camparison object with its functions and adding it to the collection of Comparisons. No need to add statements to a `switch` or `if-else if` block.

Take a look at the many posts about the `anti-if campaign` -- here's one:
https://code.joejag.com/2016/anti-if-the-missing-patterns.html

src/module/config/attributes.ts
rinickolous: True, It's likely too fast to matter. I can change it to call the function. 

As far as querying by name, this is still possible. The derived `atrributes` object has keys set as the GCS IDs. The reason the Collection doesn't just use that ID in the first place, in case I forgot to mention it in the code, is because the GCS ID is user editable and would require special update logic for when users edit it.

src/module/actor/data/gcs-character/attribute-definition.ts
mjeffw: It's not hiding it, it's unifying the behavior and logic. I'd have to see and understand the logic more to see if there should be an exception.

src/module/actor/data/gcs-character/attribute-definition.ts
mjeffw: Ok, but "kind" itself could derive the value -- unless it depends on other variables.

src/module/actor/data/gcs-character/attribute-definition.ts
mjeffw: Please.

I mean I understood it after reading a few more files, but instant understanding would be better. How about `costAdjustmentPerSizeMod`?

src/module/actor/test-actor-sheet.ts
mjeffw: I'm assuming its a simple sheet with not real considerations for layout.

src/module/data/criteria/string-criteria.ts
rinickolous: Right, that does make sense and it's actually sort of how GCS does it. The `enum` is an object you can call methods on. But the issue I have with this is that `this.compare` is a user editable field in a DataModel schema (where users would choose one of the enum options with a select box). So the type of the field generally has to be `string`, I can't have a rich object as a persistent field value.

src/module/actor/test-actor-sheet.ts
rinickolous: Yes, that's it.

src/module/actor/data/gcs-character/sheet-settings.ts
rinickolous: Sure, that would be good except we'd be brushing up against `DataModel` here, because you can't store rich objects in persistent data in Foundry. Three other options emerge if we want user-defined strength progression:
1. Keep as is but add an enum "Custom" and a second field "customProgression" which is a script field to be resolved by `ScriptResolver`.
2. Turn all of the damage progression functions into scripts for `ScriptResolver`, editable from the world settings, and able to be chosen by key from the actor.
3. Turn all of the damage progression functions into scripts for `ScriptResolver`, replace this field with a script field, edit directly on the actor (I don't like this option, it's an unnecessary amount of data on the actor and the scripts would need to be individually updated per actor if we ever need to change them)

Personally I would go with solution 2. The damage progression functionality would turn the entries in the world settings into rich objects with functions for thrust and swing, and run them for resolution.

src/module/actor/data/gcs-character/attribute-threshold.ts
rinickolous: As with my other comments about similar suggestions, I like the idea but I think it brushes up against limitations of `DataModel`. Further, in this case we can't rely on `ScriptResolver` either because `ScriptResolver` does not and cannot perform mutations of Foundry data.

I think maybe a `Feature` collection would work? Existing Features do not include the ability to multiply/divide values, but this could be added in. I'm open to suggestions.

src/module/actor/data/gcs-character/attribute-definition.ts
rinickolous: If the `type` field itself stores whether it is primary, secondary, etc. then those `isPrimary`, `isPool` fields are themselves persistent, and therefore assumed to be independent of other fields.
However, these values are necessarily not independent based on GCS logic. GCS sorts attributes into primary and secondary sets, based on whether the `base` value is just a number* (the alternative is that it's a more complex expression).

\* Can be overridden with the `placement` field.

There doesn't seem to be any good reason to make the individual fields more complex, and to make such data persistent.

Also worth noting that `TypedPseudoDocument` does not allow for the `type` field to be anything other than a string, that's already baked in.

src/module/actor/data/gcs-character/attribute-definition.ts
rinickolous: As above, it depends on other variables.

src/module/actor/data/gcs-character/sheet-settings.ts
mjeffw: For any of these multiple suggestions to turn enums into real objects, I would consider persisting the name as a string (and make it so that its obviously not the thing you interact with, renaming `DamageProgression` to `_DamageProgressionKey` or something similar). Then create a map of keys to objects, and during model instantiation, create a derived property `DamageProgression` that returns the appropriate object.

mjeffw: > Sure, that would be good except we'd be brushing up against `DataModel` here, because you can't store rich objects in persistent data in Foundry.

That seems like a major deficit in the Foundry API. I mean, the model itself is a rich object. And just what does `ObjectField` and `TypedObjectField` actually do?

PR #2557, opened 2026-02-12
Feature/foundry release

mjeffw:


rinickolous (commented):
Looks good, just need to fix that filename as Copilot pointed out

PR #2558, opened 2026-02-13
Feature: Character V1 to V2 migration

rinickolous:
## Summary

This PR implements migration of Actors of type `character` to type `characterV2`.
All items are automatically updated and a new Actor is created alongside the original (for easy comparisons). Migration can be performed by running:
```javascript
await GURPS.modules.Actor.migrateActor(actor)
```
I left in commented out code which replaces the current Actor instead of creating a new one. Migrated actors have "Migrated: " prepended to their name for easier identification though obviously this won't be the case upon release.

There are some issues I encountered during migration and applied a sort of patch fix to:
`character` V1 Reactions, Conditional Modifiers, and Weapons currently have no information contained in them regarding the Item which owns them. Given that these objects do not exist independently of Items in the new schema, they have been transplanted to a stand-in item named "The Everything Holder" (name subject to change of course). This Trait includes a description explaining what it is and asking users to re-import their actor from GCS/GCA if they wish to fix the ownership of the aforementioned objects.

Some changes to the `characterV2` schema were made for greater migration and general use convenience. Most importantly, Notes, Hit Locations, Resource Trackers, and Move Types have been turned into PseudoDocuments. Their respective fields are now ModelCollections rather than Arrays. Given that PseudoDocuments have their own update methods, methods which aim to update or add notes directly from the Actor, which treated the Note collection as an Array, have been removed.

This PR also includes some reorganization changes shared between it and my other pending feature PRs:
- Moved Item related calls from `gurps.js` to their own module
- Moved Actor related calls from `gurps.js` to their own module

Also included are small formatting changes and small bug fixes:
- Replaced `.ts` imports with `.js` so Copilot doesn't complain about imports being invalid
- Reverted a change in `anim.js` which broke querying animation files
- Removed calls to non-existent functions `_preImport` and `gurpslinkmd`
- Fixed weapon damage not being imported for V1 characters

## Potential Issues

This PR does not automatically migrate Items for non-Foundry-Item actors. We'll probably need to do that.

## Next Steps

- Test with a few characters to make sure we're not missing anything
- Implement Item migration independent of Actors
- Implement migrating data inside a compendium
- Implement automatic migration on world load
- Disable creation of new V1 Documents (disable in the UI only, Documents should remain valid and be deprecated much later)

### DIFF
--- a/src/module/actor/data/character.ts
+++ b/src/module/actor/data/character.ts
@@ -1727,7 +1727,7 @@ const characterSchema = () => {
     FP: new fields.SchemaField(poolSchema(), { required: true, nullable: false }),
     QP: new fields.SchemaField(poolSchema(), { required: true, nullable: false }),
 
-    // NOTE: This value represents "Basic Dodge", this being Math.floor(Basic Speed) + 3 + Modifiers (e.g. Enchanced
+    // NOTE: This value represents "Basic Dodge", this being Math.floor(Basic Speed) + 3 + Modifiers (e.g. Enhanced
     // Dodge). It is the base value used to get the Actual Dodge value under encumbrance.
     dodge: new fields.SchemaField({
       value: new fields.NumberField({ required: true, nullable: false, initial: 0 }),


src/module/actor/actor-components.js
```diff
@@ -726,7 +726,7 @@ export class Attack extends Named {
   /**
    * @param {string} [n1]
    * @param {string} [lvl]
-   * @param {string|Array<string>} [dmg]
```
mjeffw, line 729: Damage can actually be an array. I added this when I did the Nordlond Bestiary module. Douglas has several attacks with damage like "2d-1 cut and 1d+2 ctrl". It's not the same as followup damage - you always apply both damages on a confirmed hit. We need to care for that.

src/module/actor/legacy/actorv1-interface.ts
```diff
@@ -150,9 +150,10 @@ interface EncumbranceLevel {
   currentmovedisplay: string
 }
 
-interface Modifier {
-  description: string
-  value: number
+interface Reaction {
```
mjeffw: I suspect you've changed this name to disambiguate it from other usages of modifier. That's good. But these aren't Reactions (or rather, some are Reactions, others are Conditional Modifiers). Could we change the name to ConditionalModifier or similar? In my view, they are ALL conditional modifiers, the reaction modifiers are conditional modifiers to the reaction roll.

src/module/action/melee-attack.ts
```diff
@@ -174,8 +174,8 @@ const meleeAttackComponentSchema = () => {
     st: new fields.StringField({ required: true, nullable: false }),
     mode: new fields.StringField({ required: true, nullable: false }),
     notes: new fields.StringField({ required: true, nullable: false }),
-    weight: new fields.NumberField({ required: true, nullable: false, initial: 0 }),
-    techlevel: new fields.NumberField({ required: true, nullable: false, initial: 0 }),
+    weight: new fields.StringField({ required: true, nullable: false, initial: '' }),
```
mjeffw: What are you thinking here? That we may need the unit ('lb' vs 'kg')?

src/module/actor/migrate.ts
```diff
@@ -0,0 +1,370 @@
...
+
+    items.push({
+      _id: item._id,
+      type,
+      name: item.name,
+      system,
+    })
+  })
+
+  // ActorV1 has no concept of Reaction and Conditional Modifier ownership by items,
```
mjeffw, line 51: Long term, do you think the player should be able to enter a Conditional Modifier without an owning item? I see some utility in that (as a simpler way to add them, at least).

src/module/actor/migrate.ts
```diff
@@ -0,0 +1,370 @@
...
+  oldData: ActorV1Model
+): fields.SchemaField.CreateData<DataModel.SchemaOf<Actor.SystemOfType<'characterV2'>>> {
+  const newData: fields.SchemaField.CreateData<DataModel.SchemaOf<Actor.SystemOfType<'characterV2'>>> = {
+    attributes: oldData.attributes,
+    HP: oldData.HP,
+    FP: oldData.FP,
+    QP: oldData.QP,
+    // NOTE: dodge for characterV1 is a derived (though persistent) property, and the reported value
+    // may be adversly affected by things like encumbrance. Here, we grab the first entry in encumbrance
+    // to get our best estimate (should be accurate 99% of the time) for the base dodge value.
```
mjeffw: I think it is correct 100% of the time, as the value in array index 0 comes from GCS. If you want the current Dodge value, that is in fact derived using standard GURPS rules after calculating encumbrance bands (based on Lifting ST from, again, GCS).  I personally don't think we should persist *any* derived value, but just derive them on every data preparation cycle. Personal philosophy: don't optimize for performance until you see it is needed.

src/module/actor/migrate.ts
```diff
@@ -0,0 +1,370 @@
...
+    thrust: oldData.thrust,
+    swing: oldData.swing,
+
+    additionalresources: {
+      qnotes: oldData.additionalresources?.qnotes,
+      bodyplan: oldData.additionalresources?.bodyplan,
+      // NOTE: Ideally tracker would be a TypedObjectField
+      // instead of an ArrayField so we can updated trackers without
+      // replacing the whole array.
+      // TODO: Discuss and consider.
```
mjeffw: Agreed, or make it a pseudo collection. I don't think there's much need of discussion unless you see a problem that I don't.

src/module/actor/migrate.ts
```diff
@@ -0,0 +1,370 @@
...
+    tastesmell: oldData.tastesmell,
+    vision: oldData.vision,
+    touch: oldData.touch,
+
+    thrust: oldData.thrust,
+    swing: oldData.swing,
+
+    additionalresources: {
+      qnotes: oldData.additionalresources?.qnotes,
+      bodyplan: oldData.additionalresources?.bodyplan,
```
mjeffw: `bodyplan` should probably be at the same level as attributes in the schema. No need to keep it here.

src/module/actor/migrate.ts
```diff
@@ -0,0 +1,370 @@
...
+        value: Number(oldData.conditionalinjury.RT.value),
+        points: oldData.conditionalinjury.RT.points,
+      },
+      injury: {
+        severity: oldData.conditionalinjury.injury.severity,
+        daystoheal: oldData.conditionalinjury.injury.daystoheal,
+      },
+    },
+
+    traits: {
```
mjeffw: What's your definition of trait? I feel like we use it for several different things.

src/module/actor/migrate.ts
```diff
@@ -0,0 +1,370 @@
...
+
+      damageAccumulators: [],
+    },
+
+    hitlocationsV2: {},
+
+    moveV2: {},
+
+    // TODO: Change note into an Item or something, really shouldn't have this vestigial
+    // non-Foundry items Array present
```
mjeffw: Remove the comment since you made Notes a pseudo collection.

mjeffw: In my comments, "pseudo collection" == "pseudo document"

src/module/action/melee-attack.ts
rinickolous: Might change this back actually, since this model doesn't support custom units anyway.
And also the value is never actually usedvor displayed so it doesn't matter at all AFAICT

src/module/actor/migrate.ts
rinickolous: Might be simpler to add, but I think they should only be owned by an Item in all cases, since it's those Items that grant conditional modifiers, not other features. 

Mooks will of course be able to have them added directly.

src/module/actor/migrate.ts
rinickolous: I'll need to check if we're properly accounting for Enhanced Dodge etc. and revise this

src/module/actor/migrate.ts
rinickolous: Traits to me are Advantages and Disadvantages, so this key name is just a carryover from the original schema. I'd probably rename this to `profile` or something to match GCS

src/module/actor/migrate.ts
rinickolous: Implemented.

src/module/actor/migrate.ts
rinickolous: I reviewed this and yes, while the "current dodge" value is based on GURPS rules for encumbrance, the "Base Dodge" value is base on Math.floor(Basic Speed) + 3 + Modifiers, and since these Modifiers are not necessarily accounted for on import, a persistent Basic Dodge value remains necessary in this schema. I revised the relevatn comments to clarify on this.

src/module/actor/data/character.ts
```diff
@@ -1166,43 +1187,11 @@ class CharacterModel extends BaseActorModel<CharacterSchema> {
...
-    } else {
-      move.push({
-        mode,
-        basic: basic,
-        enhanced: enhanced ?? basic ?? 0,
-        default: isDefault ?? move.length === 0,
-      })
-    }
+    const updates = Object.fromEntries(
+      move.map(moveEntry => [`system.moveV2.${moveEntry.id}.default`, moveEntry.mode === value])
```
mjeffw, line 1191: I'm unashamed to say it took me a while to understand this code.

src/module/actor/data/character.ts
```diff
@@ -1166,43 +1187,11 @@ class CharacterModel extends BaseActorModel<CharacterSchema> {
...
-    move.forEach((moveEntry: fields.SchemaField.SourceData<MoveSchema>) => {
-      moveEntry.default = moveEntry.mode === value
-    })
-
-    await this.parent.update({ 'system.moveV2': move } as Actor.UpdateData)
-  }
-
-  /* ---------------------------------------- */
-
-  async addMoveMode({
```
mjeffw, line 1178: What happened to the `add a new Move Mode` functionality?

mjeffw (approved):

src/module/actor/data/character.ts
rinickolous: I think this function wasn't actually being called anywhere, and creation of new Move Modes (and any other PseudoDocuments) can be handled via `MoveMode.create`

PR #2563, opened 2026-02-18
Feature: Complete migration of Actors and Items to V2

rinickolous:
This change builds on the previously implemented migration functionality. List of changes:

- World Actors are migrated on world ready.
- World Items, and those in compendia, are migrated on world ready.
- Embedded Item migration within an Actor now uses the system-data pseudo-Item as a reference. Unlike Foundry items, this object always exists on both Foundry-Item and non-Foundry-Item Actors.
- Changed the Actor type `enemy` back to using the CharacterV1 model. @mjeffw if you have some reason for defining `enemy` as using the Character V2 model, let me know and we can easily change this back. It just seems like an oversight to me, but I may well be wrong.
- Changed all places in the code referring to `system.traits` to use `system.profile`, and the same for `system.additionalresources.bodyplan` -> `system.bodyplan` (with the exception of deprecated code). This is to match the schema changes from the previous PR.
- Made `tsc-alias` run even if `tsc` shows errors. This was changed to ensure that even if there are some type errors present in the code, it can still be opened in Foundry (for debugging, etc.)
- Added GCA 5 direct import functionality to changelog

### DIFF
--- a/src/module/action/melee-attack.ts
+++ b/src/module/action/melee-attack.ts
@@ -47,6 +47,9 @@ class MeleeAttackModel extends BaseAction<MeleeAttackSchema> {
    * Prepare the level of this skill based on an OTF formula.
    */
   #prepareLevelsFromOtf(): void {
+    // Do not prepare levels if the item is not owned
+    if (!this.item.isOwned) return
+
     let otf = this.component.otf
 
     if (otf === '') {
--- a/src/module/action/ranged-attack.ts
+++ b/src/module/action/ranged-attack.ts
@@ -45,6 +45,9 @@ class RangedAttackModel extends BaseAction<RangedAttackSchema> {
    * Prepare the level of this skill based on an OTF formula.
    */
   #prepareLevelsFromOtf(): void {
+    // Do not prepare levels if the item is not owned
+    if (!this.item.isOwned) return
+
     let otf = this.component.otf
 
     if (otf === '') {


package.json
```diff
@@ -52,7 +52,7 @@
     "typecheck": "tsc --noEmit",
     "verify": "npm run format:check && npm run lint && npm run typecheck && npm run test",
     "watch": "node build/watch.mjs",
-    "watch:code": "tsc-watch -p tsconfig.json --onSuccess \"tsc-alias -p tsconfig.json\"",
+    "watch:code": "tsc-watch -p tsconfig.json --onSuccess \"tsc-alias -p tsconfig.json\" --onFailure \"tsc-alias -p tsconfig.json\"",
```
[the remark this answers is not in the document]
rinickolous: Keeping as is

src/module/actor/migrate.ts
```diff
@@ -156,38 +233,18 @@ async function migrateActor(actor: Actor.Implementation): Promise<Actor.OfType<'
...
     type: 'characterV2',
     img: actor.img,
-    name: 'Migrated: ' + actor.name,
+    name: actor.name,
     system: migrateActorSystem(actor.system),
     items,
   }
 
-  const newActor = Actor.create(createData) as unknown as Actor.OfType<'characterV2'>
+  const newActor = (await actor.update(createData, { recursive: false })) as unknown as Actor.OfType<'characterV2'>
```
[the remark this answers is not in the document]
rinickolous: Not a real issue.

src/module/item/migrate.ts
```diff
@@ -13,7 +13,126 @@ type CreateDataOf<Model extends DataModel.Any> = fields.SchemaField.CreateData<D
...
+      for (const item of items) {
+        await migrateItem(item)
+        updateProgress += updateStep
+        warning.update({ pct: updateProgress })
+      }
+
+      for (const pack of packs) {
+        await migrateItemCompendium(pack)
+        updateProgress += pack.index.size * updateStep
+        warning.update({ pct: updateProgress })
```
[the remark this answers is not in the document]
rinickolous: The migration process is so fast that the progress bar is only expected to appear for a second or so. I don't think this big jump matters.

src/module/actor/migrate.ts
```diff
@@ -10,14 +9,62 @@ import { MoveModeV2 } from './data/move-mode.js'
...
+      for (const actor of actors) {
+        await migrateActor(actor)
+        updateProgress += updateStep
+        warning.update({ pct: updateProgress })
+      }
+
+      for (const pack of packs) {
+        await migrateActorCompendium(pack)
+        updateProgress += pack.index.size * updateStep
+        warning.update({ pct: updateProgress })
```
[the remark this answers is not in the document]
rinickolous: The migration process is so fast that the progress bar is only expected to appear for a second or so. I don't think this big jump matters.

src/module/actor/actor-sheet.js
```diff
@@ -1776,9 +1776,9 @@ export class GurpsActorSheet extends foundry.appv1.sheets.ActorSheet {
     event.preventDefault()
 
     switch (this.actor.type) {
```
mjeffw, line 1778: I probably just missed this on the last review, but the type check method I created in Actor (isNewType or something like that) ought to be used instead as that centralizes the logic for whether you'd choose a V1 or V2 actor type.

src/module/util/miscellaneous-settings.js
```diff
@@ -81,6 +83,17 @@ export function initializeSettings() {
   Hooks.once('init', async function () {
     // Game Aid Information Settings ----
 
+    // Dev Mode
```
mjeffw: I'd really like to deprecate this file entirely. New settings should be added to modules or a generic placeholder location. I suggest we have something like `src/module/util/dev` which is a development utility module, and move things like the `debugger.js` to that directory as well.

rinickolous: Great idea.

src/module/actor/actor-sheet.js
rinickolous: The accessor `isNewActorType` is private so I can't use it outside of the class. Is that intentional? I can't actually use `isNewActorType` in the place you've put this comment down in, for example.

mjeffw: It can be made public.

PR #2566, opened 2026-02-23
Feature: GCS Compendium Importer

rinickolous:
This PR updates the GCS Importer functionality, expanding it to allow importing GCS Libraries for Traits, Skills, Spells, and Equipment. It also amends some weapon importing functionality which was previously missing: Automatic conversion of weapon skill defaults to OTF strings, allowing for dynamic weapon level calculation.

This PR extends the Compendium Directory Application and adds an action to it which brings up the prompt rather than dynamically adding the button to the sidebar tab every time it renders.

I had a look at the potential for extending this to GCA 5 libraries too, but the file format of these library files is the old proprietary format rather than the XML format GCA5 character sheets use.

This PR also fixes a bug I happened upon during implementation: Weapons attempt to resolve their level during data preparation if they are not owned, but this value depends on them being owned. I made it skip the level resolution functionality if the weapon is not owned.

### DIFF
--- a/src/module/importer/gcs-importer/schema/character.ts
+++ b/src/module/importer/gcs-importer/schema/character.ts
@@ -109,9 +109,10 @@ class GcsCharacter extends GcsElement<GcsCharacterModel> {
   /* ---------------------------------------- */
 
   get allNotes(): GcsNote[] {
-    const notes = this.notes ?? []
+    const notes = []
 
-    for (const note of notes) {
+    for (const note of this.notes ?? []) {
+      notes.push(note)
       notes.push(...note.allChildItems)
     }
 


src/module/action/melee-attack.ts
```diff
@@ -47,6 +47,8 @@ class MeleeAttackModel extends BaseAction<MeleeAttackSchema> {
    * Prepare the level of this skill based on an OTF formula.
    */
   #prepareLevelsFromOtf(): void {
```
mjeffw: Could this work for both melee-attack and ranged-attack, and be moved to BaseAction?

src/module/action/ranged-attack.ts
```diff
@@ -45,6 +45,8 @@ class RangedAttackModel extends BaseAction<RangedAttackSchema> {
    * Prepare the level of this skill based on an OTF formula.
    */
   #prepareLevelsFromOtf(): void {
```
mjeffw: Same comment as for melee-attack.

src/module/compendium/compendium-directory.ts
mjeffw, line 1: I'd strongly prefer having a compendium module instead of a catch-all "applications" module. Or maybe something that looks like it explicitly extends the Foundry UI.

src/module/compendium/index.ts
mjeffw, line 1: Compendium or Foundry extensions  module.

src/module/action/melee-attack.ts
rinickolous: Not quite, but removed the duplication.

src/module/importer/gcs-importer/importer.ts
```diff
@@ -117,11 +165,126 @@ class GcsImporter {
...
+
+  #existingItemId(itemData: Item.CreateData, existingIds: { _id: string | null; importid: string }[]): string | null {
+    const getImportId = (data: Item.CreateData) => {
+      if (this.#itemDataIsOfType(data, 'featureV2')) return data.system?.fea?.importid
+      if (this.#itemDataIsOfType(data, 'skillV2')) return data.system?.ski?.importid
+      if (this.#itemDataIsOfType(data, 'spellV2')) return data.system?.spl?.importid
+      if (this.#itemDataIsOfType(data, 'equipmentV2')) return data.system?.eqt?.importid
+
+      return null
+    }
```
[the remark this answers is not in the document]
rinickolous: Will be resolved by other PR anyway

src/module/importer/gcs-importer/importer.ts
```diff
@@ -117,11 +165,126 @@ class GcsImporter {
...
+      console.error("GURPS | Foundry game.packs is undefined. Can't import compendium.")
+
+      return null
+    }
+
+    let pack = game.packs.get(`world.${name}`) as foundry.documents.collections.CompendiumCollection<'Item'> | undefined
+
+    const existingItems: { _id: string | null; importid: string }[] = pack
+      ? (await pack.getDocuments()).map(item => {
+          return { _id: item._id || null, importid: item.system.component.importid || '' }
```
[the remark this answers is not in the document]
rinickolous: This will be resolved by #2569

PR #2569, opened 2026-02-26
Refactor: Item <-> Item Component Merge

rinickolous:
NOTE: This PR is dependent on #2566 . Let's deal with that one before this one.

This PR merges all DataModel classes which were previously split into a parent and component class (.e.g. `TraitModel` and `TraitComponentModel`).

It also does the following:
- Converts Reaction Bonuses and Conditional Modifiers into PseudoDocuments: These were previously Arrays of schemas, which would have made adding, modifying, and removing reactions annoying as specific update methods would be necessary to manage the arrays containing these objects.
- Somewhat improves on my merge of some functionality of MeleeWeapon and RangedWeapon in #2566 . I prefer this implementation to the other and 
- Removes what appears to be legacy migration functionality targeting V1 characters in order to migrate some objects to forms more closely matching the V2 schema.
- Fixes a bug in `GurpsActorV2#_preUpdate` which would cause DataModel-specific preUpdate actions to not run.
- Removes some unused functions from `GurpsActorV2`, pertaining to Reactions and Conditional Modifiers above.
- Suppresses unneeded console messages (e.g. migration messages when a migration doesn't do anything)
- Duplicates the body plan overwrite prompt for imports from GCA (already present for GCS imports)
- Fixes a bug that would cause hit locations to duplicate on re-import.
- Adds explanatory documentation to various fields in Items and Attacks.
- Removes some unused / vestigial fields (e.g. weight and TL for melee weapons, where weapons do not have a weight or TL independent of their parent Item)
- Carries over some fixes for TypedPseudoDocument (better return resolving for the `create` method, etc.) from my other PRs
- Removes the pre-update tests file, which targets legacy functionality which is removed as above.

### DIFF
--- a/src/module/action/ranged-attack.ts
+++ b/src/module/action/ranged-attack.ts
@@ -38,7 +38,7 @@ class RangedAttackModel extends BaseAttack<RangedAttackSchema> {
     if (game.settings?.get(GURPS.SYSTEM_NAME, Settings.SETTING_CONVERT_RANGED) === false) return
 
     const range = this.range
-    // Match the range format, e.g., "x2", "x20", "x30"
+    // Match the range format, e.g., "×2", "x20", "X30"
     const matchSingle = range.match(/^\s*[×xX]([\d.]+)\s*$/)
     const matchMultiple = range.match(/^\s*[×xX]([\d.]+)\s*-\s*[×xX]([\d.]+)\s*$/)
 


src/module/importer/gca-importer/importer.ts
```diff
@@ -559,13 +678,6 @@ Portrait will not be imported.`
       rcl: weapon.charrcl ?? '',
       halfd: weapon.charrangehalfdam ?? '',
```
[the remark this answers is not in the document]
rinickolous: These can resolve to default values.

src/module/importer/gca-importer/importer.ts
```diff
@@ -529,26 +654,20 @@ Portrait will not be imported.`
       parry: weapon.charparry ?? '',
       block: weapon.charblockscore ?? '',
```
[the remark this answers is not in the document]
rinickolous: These can resolve to default values.

src/module/importer/gcs-importer/importer.ts
```diff
@@ -672,27 +724,21 @@ Portrait will not be imported.`
       block: weapon.calc?.block || weapon.block,
       blockbonus,
```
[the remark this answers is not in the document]
rinickolous: These can resolve to default values.

src/module/importer/gcs-importer/importer.ts
```diff
@@ -703,16 +749,9 @@ Portrait will not be imported.`
       range: weapon.calc?.range || weapon.range,
       rcl: weapon.calc?.recoil || weapon.recoil,
       halfd,
-      rate_of_fire: weapon.calc?.rate_of_fire || weapon.rate_of_fire,
+      rateOfFire: weapon.calc?.rate_of_fire || weapon.rate_of_fire,
       bulk: weapon.calc?.bulk || weapon.bulk || '0',
```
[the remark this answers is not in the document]
rinickolous: These can resolve to default values.

src/module/action/melee-attack.ts
```diff
@@ -118,36 +59,36 @@ class MeleeAttackModel extends BaseAction<MeleeAttackSchema> {
...
-        if (this.component.name.match(makeRegexPatternFrom(bonus.name as string, false))) {
-          this.component.level += bonus.mod as number
+        if (this.name && this.name.match(makeRegexPatternFrom(bonus.name as string, false))) {
+          this.level += bonus.mod as number
 
           // Handle parry with a suffix such as "F"
-          if (!isNaN(parseInt(this.component.parry))) {
-            const parryLevel = parseInt(this.component.parry)
-            const parrySuffix = this.component.parry.replace(parryLevel.toString(), '').trim()
+          if (!isNaN(parseInt(this.parry))) {
```
mjeffw, line 83: I feel like we shouldn't be dependent on parry (and other similar values) being either a number or a string, or even a string that sometimes has `parseInt` compatible value and other times, not. In our model, Parry should be an object with a level and flags to indicate Fencing parry or "parry causes the weapon to be unready".  Then we simply display it as "12F" or "9U" or whatever.

src/module/action/melee-attack.ts
```diff
@@ -159,71 +100,37 @@ class MeleeAttackModel extends BaseAction<MeleeAttackSchema> {
...
-    mode: new fields.StringField({ required: true, nullable: false }),
-    notes: new fields.StringField({ required: true, nullable: false }),
-    weight: new fields.NumberField({ required: true, nullable: false, initial: 0 }),
-    techlevel: new fields.StringField({ required: true, nullable: false, initial: '' }),
-    cost: new fields.StringField({ required: true, nullable: false }),
+    /** The reach of this attack, e.g. "C", "C,1", "1,2", etc. */
     reach: new fields.StringField({ required: true, nullable: false }),
+
+    /** The parry value of this attack, e.g. "3", "3F",  etc. */
     parry: new fields.StringField({ required: true, nullable: false }),
```
mjeffw, line 107: See my note about parry, above.

src/module/action/ranged-attack.ts
```diff
@@ -104,18 +37,18 @@ class RangedAttackModel extends BaseAction<RangedAttackSchema> {
   convertRanges(st: number): void {
     if (game.settings?.get(GURPS.SYSTEM_NAME, Settings.SETTING_CONVERT_RANGED) === false) return
 
-    const range = this.component.range
+    const range = this.range
     // Match the range format, e.g., "x2", "x20", "x30"
```
mjeffw: > `// Match the range format, e.g., "x2", "x20", "x30"`

A bit of a misleading comment -- it should probably be `"×2", "x20", "X30"`.

Also, is `"*2"` a valid format?

mjeffw (changes requested):

src/module/action/melee-attack.ts
rinickolous: Yes, that's what GCS does too. I'd be happy to move to a format more closely resembling that and to implement any necessary migration steps for old data.

mjeffw: Go for it. Let's do it in  another PR, though.

src/module/action/ranged-attack.ts
rinickolous: Doesn't look like my comment but I'll get that changed.

rinickolous: I think we'll go without `*` as multiplier. Neither GCS nor GCA use it in their formatting, and what we have is good enough. Comment changed.

PR #2571, opened 2026-02-27
Feature: Complex fields for Attacks

### LINKED ISSUE #2570: Update complex weapon stats to objects rather than parsed strings or numbers
I feel like we shouldn't be dependent on parry (and other similar values) being either a number or a string, or even a string that sometimes has `parseInt` compatible value and other times, not. In our model, Parry should be an object with a level and flags to indicate Fencing parry or "parry causes the weapon to be unready".  Then we simply display it as "12F" or "9U" or whatever.

_Originally posted by @mjeffw in https://github.com/crnormand/gurps/pull/2569#discussion_r2860992203_

rinickolous:
This PR transforms existing melee and ranged attack fields from primitives to objects closely mimicking those found in GCS.
Old data migration has been amended to transform the old field types into the new ones, as have the GCA and GCS importers.
The parsing functions also have Jest tests attached.

~~One thing I haven't implemented at this stage is the possibility of modifying the objects by passing through a string which is then parsed using the parsing function. I think the "correct" way to do this would be defining `DataModel` classes for each of the fields and using the `_cast` method. I plan on doing that anyway for the GCS versions of the same actions, so it may be worth doing now, but I'll leave that up to you to decide.~~

Edit: The `_cast` methods are now implemented, since I realized it doesn't actually require converting the fields to `DataModel` subclasses.

Resolves #2570

### DIFF
--- a/src/module/resource-tracker/index.ts
+++ b/src/module/resource-tracker/index.ts
@@ -4,7 +4,7 @@ import { migrateTrackerInstanceToV2 } from './migration.js'
 import { migrate } from './migration.js'
 import { initializeSettings } from './settings.js'
 import { IResourceTracker, IResourceTrackerTemplate } from './types.js'
-import { ResourceTrackerManagerV2 } from './ui/resource-tracker-manager-v2.ts'
+import { ResourceTrackerManagerV2 } from './ui/resource-tracker-manager-v2.js'
 import { updateResourceTracker } from './ui/update-resource-tracker.js'
 
 function init() {
@@ -73,7 +73,7 @@ export const ResourceTrackerModule: ResourceTrackerModule = {
 
 export type { IResourceTrackerThreshold, IResourceTracker, IResourceTrackerTemplate } from './types.js'
 
-export { ResourceTrackerManagerV2 } from './ui/resource-tracker-manager-v2.ts'
+export { ResourceTrackerManagerV2 } from './ui/resource-tracker-manager-v2.js'
 export type { ResourceTrackerSchema, ResourceTrackerTemplateSchema } from './resource-tracker.js'
 export { TrackerInstance, ResourceTrackerTemplate } from './resource-tracker.js'
 export { OperatorFunctions, ComparisonFunctions } from './types.js'
--- a/src/module/resource-tracker/settings.ts
+++ b/src/module/resource-tracker/settings.ts
@@ -1,5 +1,5 @@
 import { IResourceTrackerTemplate, SETTING_TRACKER_EDITOR, SETTING_TRACKER_TEMPLATES } from './types.js'
-import { ResourceTrackerManagerV2 } from './ui/resource-tracker-manager-v2.ts'
+import { ResourceTrackerManagerV2 } from './ui/resource-tracker-manager-v2.js'
 
 type ResourceTrackerTemplateMap = Record<string, IResourceTrackerTemplate>
 type ResourceTrackerTemplateMapConstructor = ObjectConstructor & {
--- a/src/module/resource-tracker/ui/resource-tracker-editor-v2.ts
+++ b/src/module/resource-tracker/ui/resource-tracker-editor-v2.ts
@@ -3,7 +3,7 @@ import { DeepPartial } from 'fvtt-types/utils'
 import { TrackerInstance } from '../resource-tracker.js'
 import { IResourceTracker, IResourceTrackerTemplate, TrackerComparators, TrackerOperators } from '../types.js'
 
-import { ResourceTrackerManagerV2 } from './resource-tracker-manager-v2.ts'
+import { ResourceTrackerManagerV2 } from './resource-tracker-manager-v2.js'
 
 type ResourceTrackerEditorV2Options = DeepPartial<foundry.applications.api.ApplicationV2.Configuration> & {
   onUpdate?: (trackerData: IResourceTracker) => Promise<void> | void


src/module/action/melee-attack.ts
```diff
@@ -1,27 +1,83 @@
...
+  static override cleanData(source?: AnyMutableObject, options?: fields.DataField.CleanOptions): AnyMutableObject {
+    source = super.cleanData(source, options)
+
+    // Clean the parry field to ensure it is normalized.
+    // If parrying is disabled, the other parry fields are also reset to their default values.
+    if ('parry' in source && typeof source.parry === 'object' && source.parry !== null) {
+      if ('canParry' in source.parry && typeof source.parry.canParry === 'boolean' && !source.parry.canParry) {
+        ;(source.parry as AnyMutableObject).fencing = false
+        ;(source.parry as AnyMutableObject).unbalanced = false
+        ;(source.parry as AnyMutableObject).modifier = 0
```
mjeffw: Get rid of any line that starts with `;(...` -- For instance in this case, do:
```js
const parry = source.parry as AnyMutableObject
parry.fencing = false
parry.unbalanced = false
...
```
Apply the same across the codebase -- is this something we can set a lint rule for?

src/module/item/migrate.ts
```diff
@@ -179,18 +191,20 @@ function migrateBaseItemSystem(oldData: OldItemData, parentId: string | null): N
   }
 
   if (oldData.melee) {
-    Object.values(oldData.melee).forEach(action => {
-      const id = foundry.utils.randomID()
+    ;(Object.values(oldData.melee) as Melee[]).forEach(action => {
```
mjeffw: Same as a previous comment -- split up the lines so we don't start a line with `;(...`.

```
const melee = Object.values(oldData.melee) as Melee[]
melee.forEach(actions => { ...
```

rinickolous:
> > One thing I haven't implemented at this stage is the possibility of modifying the objects by passing through a string which is then parsed using the parsing function. I think the "correct" way to do this would be defining `DataModel` classes for each of the fields and using the `_cast` method. I plan on doing that anyway for the GCS versions of the same actions, so it may be worth doing now, but I'll leave that up to you to decide.
> 
> Can you give me an example of what you're talking about? Are you talking about portions of the attacks (e.g., Damage) or the whole attack line (as might be entered via the `/mook` editor, for example), or something else?
> 
> And how would the `_cast` method be used (and should it be used -- it is a protected method)?

Sure. The `foundry.data.DataField` class has a `_cast` method which it runs as pre-processing of data assigned to the field to enforce the primitive type. So `NumberField` will take whatever input it receives and produce a number, which is then passed on for further cleaning via `_cleanType` which might do things like enforce minimum and maximum values, etc.

For existing implementation of this method, look at `LengthField` and `WeightField`. `Length` and `Weight` are DataModels with `value` and `unit` fields, but the user can pass a number or a string like `5 lb` to the field, and the `_cast` method then calls the string/number parser to get an object of the right type.

I was actually incorrect in my initial assumption (going from memory at the time) that `_cast` is a method of `DataModel`. It's actually a method of `DataField` which means we wouldn't have to turn these weapon objects into class instances (good, simplifies things). We would just need to add a subclass of `SchemaField` for these fields and override `_cast` in them, passing a string value to the parser.

As for practical applications like using this in the mook generator, that would certainly be a good application of the above. We could, for example, do something like check parts of an attack line for the basic shape required for a given field, and then pass that to the parser for that field.

I haven't touched the `damage` field in this PR mostly to avoid brushing up against your existing implementation, but the logical endpoint of these changes includes doing so for the `damage` field too.

src/module/action/parse-weapon.ts
mjeffw, line 1: In some cases maybe it makes sense to organize utilities to mirror the module directory structure. So consider moving this to `src/module/util/action`.

src/module/action/legacy/meleev1.ts
```diff
@@ -36,9 +36,10 @@ class MeleeV1 {
     return []
   }
 
-  get cost() {
-    return this.meleeV2.cost
-  }
+  // not sure what cost is supposed to do
```
mjeffw: You can remove it.

src/module/action/ranged-attack.ts
```diff
@@ -17,37 +42,148 @@ class RangedAttackModel extends BaseAttack<RangedAttackSchema> {
     return ActionType.RangedAttack
   }
 
+  /* ---------------------------------------- */
+
+  static override LOCALIZATION_PREFIXES: string[] = ['GURPS.Action.RangedAttack']
```
mjeffw: We are preferring camelCase for i18n tags: `GURPS.action.rangedAttack`.

src/module/action/melee-attack.ts
```diff
@@ -1,28 +1,87 @@
...
+
+  /* ---------------------------------------- */
+
   static override get TYPE(): string {
     return ActionType.MeleeAttack
   }
 
+  /* ---------------------------------------- */
+
+  static override LOCALIZATION_PREFIXES: string[] = ['GURPS.Action.MeleeAttack']
```
mjeffw: camelCase = `GURPS.action.meleeAttack`

src/module/action/parse-weapon.ts
mjeffw: Actually, after reviewing all these files, why not just move it to `src/module/action`? The `action` module is the only place they are used. AND -- these methods depend heavily on the code in `action` and so they should be moved there.

I think in general, the code in `utils` should be mostly independent of code in a module.

src/module/action/melee-attack.ts
```diff
@@ -1,28 +1,87 @@
...
+
+  /* ---------------------------------------- */
+
   static override defineSchema(): MeleeAttackSchema {
     return Object.assign(super.defineSchema(), meleeAttackSchema())
   }
 
   /* ---------------------------------------- */
 
+  static override cleanData(source?: AnyMutableObject, options?: fields.DataField.CleanOptions): AnyMutableObject {
```
mjeffw, line 27: Good method to unit test.

src/module/action/melee-attack.ts
```diff
@@ -31,27 +90,38 @@ class MeleeAttackModel extends BaseAttack<MeleeAttackSchema> {
...
+    this.parryLevel = this.parry.canParry ? Math.floor(this.level / 2) + 3 + this.parry.modifier : 0
 
-      this.parry = `${3 + Math.floor(this.level / 2) + this.parrybonus}${parrySuffix}`
-    }
+    this.blockLevel = this.block.canBlock ? Math.floor(this.level / 2) + 3 + this.block.modifier : 0
+  }
+
+  /* ---------------------------------------- */
+
+  #prepareDisplayValues(): void {
```
mjeffw, line 114: Another place for unit tests.

src/module/action/ranged-attack.ts
```diff
@@ -17,37 +42,148 @@ class RangedAttackModel extends BaseAttack<RangedAttackSchema> {
...
+  /* ---------------------------------------- */
+
   override prepareDerivedData(): void {
     super.prepareDerivedData()
+    this.#prepareMusclePoweredRange()
+  }
+
+  /* ---------------------------------------- */
+
+  #prepareDisplayValues(): void {
```
mjeffw, line 206: Unit tests?

src/module/action/ranged-attack.ts
mjeffw, line 1: Consider unit testing all methods that format data for display, do calculations, or parse strings into persistent data.

Not for this PR, but I would also like to think about how to move calculations based on the rules into the `@rules` code.

static/lang/de.json
```diff
@@ -11,6 +11,94 @@
     "abbrevReference": "Ref",
     "acc": "ZG",
     "accForWeapon": "ZG für {weapon}",
+    "action": {
+      "baseAction": {
+        "FIELDS": {
```
mjeffw, line 16: I'm not sure I think the "FIELDS" grouping is necessary or even a good name.

static/lang/en.json
```diff
@@ -11,6 +11,94 @@
...
+          "consumeAction": "Consume Action",
+          "damage": "Damage",
+          "extraAttacks": "Extra Attacks",
+          "import": "Base Level",
+          "itemModifiers": "Item Modifiers",
+          "mode": "Mode",
+          "modifierTags": "Modifier Tags",
+          "notes": "Notes",
+          "otf": "OTF Formula",
+          "st": "ST"
```
mjeffw, line 33: Here's something I struggle with: "ST" is used throughout the system, do we really want to redefine it in every module? Or would we want to set up a "miscellaneous" or "standard" or "system" group of language tags and refer to them throughout the modules? I would prefer to do the latter.

mjeffw: Same for "OTF Formula", and many other tags.

static/lang/de.json
rinickolous: I'm using it because Foundry automatically recognises fields under the FIELDS key when LOCALIZATION_PREFIXES is defined on the model. This lets us use the formGroup and formField helpers and define the schema fields for a model without explicitly defining the labels in all of these places

static/lang/en.json
rinickolous: I agree, I think we should go through the lang files and properly consolidate things by category. Some of the redefined keys here are just temporary and there for convenience. 

But I think such a consolidation belongs in its own PR

src/module/action/ranged-attack.ts
rinickolous: Sure. I already have unit tests for parsing strings into persistent data, but will put in tests for going the other way too.

src/module/action/ranged-attack.ts
```diff
@@ -1,53 +1,329 @@
...
   }
 
   /* ---------------------------------------- */
 
-  convertRanges(st: number): void {
-    if (game.settings?.get(GURPS.SYSTEM_NAME, Settings.SETTING_CONVERT_RANGED) === false) return
+  #prepareMusclePoweredRange(): void {
+    if (game.settings?.get(GURPS.SYSTEM_NAME, 'convert-ranged') === false) return
+    if (!this.item.isOwned) return
+
```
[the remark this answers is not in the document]
rinickolous: we're keeping this as is at least for now to avoid circular imports which break unit testing.

static/lang/de.json
mjeffw: Cool. LGTM

PR #2572, opened 2026-03-04
Feature/select energy pool

mjeffw:
Add the ability to prompt the user for which energy pool to use when processing a "*costs X " OTF.

Currently OTF allows for terms like "+1d-1 cr *per 1 FP" to describe Fireball spell-like damage. When the OTF is executed, the damage is rolled as Fatigue is reduced by 1.

For standard GURPS spell-casting, the energy may come from Fatigue or Hit Points or from an Energy Reserve (either as an advantage or an item). ("*per 1 HP" would pull from HP, "*per 1 tr(Energy Reserve)" or "*per 1 tr0001" would pull from a resource tracker.)

To prompt the user for which to use, the OTF should NOT specify a known pool (for example "+1d burn *per 1 point"). When processing, GGA will now prompt the user for either FP, HP, or a resource tracker name.

In addition, this PR:

- Moves a function for processing OTF *Costs and *Max into a src/module/otf directory, in preparation for an OTF module being created.
- Defines a new function for doing World Settings migrations (module/utils/migration).
- Updated the importer and resource-tracker modules to use this new settings migration function.
- Fixed some code to add ResourceTrackers in Modern sheet and added the default trackers in the import process.
- Updated the Resource Tracker UI to ApplicationV2 and Modern UI styling.
- Added tooltips to the Modern sheet HP and FP trackers.
- Added instantiation of default trackers in both GCS and GCA importers.
- Reorganizes some tests and the @rules directory for clarity.
- Moves some code and related tests from @module to @rules.

### DIFF
--- a/src/module/actor/modern/dialog-crud-handler.ts
+++ b/src/module/actor/modern/dialog-crud-handler.ts
@@ -227,6 +227,7 @@ export function bindTrackerActions(html: HTMLElement, actor: Actor.Implementatio
         }
       )
 
+      // TODO: Make DR modifiers derived rather than persisted.
       if (confirmed) {
         await actor.refreshDR()
       }
--- a/src/module/actor/modern/sheet.ts
+++ b/src/module/actor/modern/sheet.ts
@@ -99,11 +99,6 @@ export class GurpsActorModernSheet extends SheetBase {
       importActor: GurpsActorModernSheet.#onImportActor,
       editQuickNotes: GurpsActorModernSheet.#onEditQuickNotes,
       editMoveMode: GurpsActorModernSheet.#onEditMoveMode,
-      // editTracker: GurpsActorModernSheet.#onEditTracker,
-      // deleteTracker: GurpsActorModernSheet.#onDeleteTracker,
-      // decreaseTracker: GurpsActorModernSheet.#onUpdateTrackerValue,
-      // increaseTracker: GurpsActorModernSheet.#onUpdateTrackerValue,
-      // resetTracker: GurpsActorModernSheet.#onUpdateTrackerValue,
     },
   }
 
--- a/src/module/resource-tracker/ui/resource-tracker-manager-v2.ts
+++ b/src/module/resource-tracker/ui/resource-tracker-manager-v2.ts
@@ -1,9 +1,9 @@
 import { DeepPartial } from 'fvtt-types/utils'
 
-import { TrackerComparators, TrackerOperators } from '../types.ts'
-import { IResourceTracker, IResourceTrackerTemplate, SETTING_TRACKER_TEMPLATES } from '../types.ts'
+import { TrackerComparators, TrackerOperators } from '../types.js'
+import { IResourceTracker, IResourceTrackerTemplate, SETTING_TRACKER_TEMPLATES } from '../types.js'
 
-import { ResourceTrackerEditorV2 } from './resource-tracker-editor-v2.ts'
+import { ResourceTrackerEditorV2 } from './resource-tracker-editor-v2.js'
 
 type ResourceTrackerManagerV2Context = foundry.applications.api.ApplicationV2.RenderContext & {
   templates: Record<string, IResourceTrackerTemplate>
--- a/src/module/util/contrast-color.ts
+++ b/src/module/util/contrast-color.ts
@@ -1,5 +1,3 @@
-'use strict'
-
 /**
  * Determine the luminescence of the provided background color, and return either a light or dark foreground color to
  * contrast with it.


build/cpx.js
rinickolous, line 1: Looks like a merge/rebase artifact. File shouldn't exist.

src/module/resource-tracker/ui/update-resource-tracker.ts
```diff
@@ -0,0 +1,19 @@
...
+  const trackerData = JSON.parse(JSON.stringify(tracker)) as IResourceTracker
+
+  interface TrackerEditorOptions {
+    onUpdate: (editedTracker: IResourceTracker) => Promise<void>
+  }
+
+  const dialog = new GURPS.modules.ResourceTracker.TrackerEditorV2(trackerData, {
+    onUpdate: async (editedTracker: IResourceTracker): Promise<void> => {
+      await tracker.update(editedTracker as any)
+      await actor.render()
```
rinickolous: A PseudoDocument update _should_ already trigger an update (and therefore an app refresh) for the affected actor, so I'm not sure whether this additional render call is necessary.

src/module/otf/description-utilities.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    if (system.HP) costs[`system.HP.value`] = `${game.i18n!.localize('GURPS.HP')} (${system.HP.value})`
+    if (system.FP) costs[`system.FP.value`] = `${game.i18n!.localize('GURPS.FP')} (${system.FP.value})`
+
+    for (const [key, tracker] of Object.entries((system.additionalresources?.tracker ?? {}) as Record<string, any>)) {
+      const trackerName = (tracker?.name ?? '').toString().trim()
+      const label = trackerName !== '' ? trackerName : `Tracker ${key}`
+      const value = tracker?.value ?? ''
+
+      costs[`system.additionalresources.tracker.${key}.value`] = `${label} (${value})`
```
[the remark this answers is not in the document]
rinickolous: I agree with this but will leave implementation up to you.

src/module/resource-tracker/resource-tracker.ts
```diff
@@ -38,13 +49,118 @@ class TrackerInstance extends PseudoDocument<ResourceTrackerSchema> {
...
+
+      const updates: Partial<TrackerInstance> = {}
+
+      updates.max = value
+      updates.value = this.isAccumulator ? this.min : value
+      updates.initialValue = null
+
+      await this.update(updates)
+    }
+  }
```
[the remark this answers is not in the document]
rinickolous: I agree. We should never be calling `this.update` from `prepareDerivedData` or any other data preparation method. Setting an initial value for fields of a DataModel belongs in `_initialize` or some other related function.

src/module/util/hitpoints.js
```diff
@@ -60,12 +60,12 @@ const hpConditions = {
...
     breakpoint: _ => Number.MAX_SAFE_INTEGER,
-    label: 'GURPS.normal',
+    label: 'GURPS.Healthy',
     style: 'normal',
   },
   REELING: {
     breakpoint: FP => Math.ceil(FP.max / 3) - 1,
-    label: 'GURPS.status.tired',
+    label: 'GURPS.status.Tired',
     style: 'tired',
```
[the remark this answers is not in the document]
rinickolous: I agree.

src/module/util/moustachewax.js
```diff
@@ -152,6 +145,47 @@ export default function () {
     return Math.round((1 - currentHP / maxHP) * 100)
   })
 
+  Handlebars.registerHelper('hp-thresholds', function (maxHP) {
```
rinickolous: Why are we using a Handlebars helper for this functionality?
I believe it would be cleaner to prepare an object with the values presented here in context preparation, and use a template or maybe just a loop to do the actual rendering.

static/templates/actor/modern/partials/ms-resources.hbs
```diff
@@ -16,9 +16,9 @@
       <i class='fas fa-rotate-left'></i>
     </button>
   </div>
-  <div class='ms-resource ms-fp'>
+  <div class='ms-resource ms-fp' data-tooltip='{{{fp-thresholds system.FP.max}}}'>
     {{#if system.conditions.exhausted}}<span class='ms-condition-badge' title='{{localize "GURPS.FP.tired"}}'>T</span>{{/if}}
-    <div class='ms-resource-display' data-edit-target='fp' title='{{localize "GURPS.attributesFPNAME"}}'>
+    <div class='ms-resource-display' data-edit-target='fp'>
```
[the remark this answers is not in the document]
rinickolous: I agree.

src/module/resource-tracker/resource-tracker.ts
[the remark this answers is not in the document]
mjeffw: @rinickolous This code needs to be executed on any update to the initialValue field of the object. The initialization needs to occur after the containing actor, if any, has been initialized and all modifiers applied. For example, for Control Points, the tracker needs to be initialized from the actor's Lifting ST, which will only be known after any advantages that modify the character's Lifting ST has been applied. Where is the best place to put this initialization code?

Actually, we can't even guarantee that the initialValue field will be updated during CRUD processing (pseudodoc.update) because it may be present at object instantiation. The initialization code needs to run whenever we detect that initialValue contains a value.

Until I can figure out the correct place to put this initialization, I'm going to simply change the `prepareDerivedData` method back to synchronous and remove the `await` on the call to update the resource tracker pseudo document. I think this may cause the Resource Tracker to be initialized twice, but that seems harmless.

src/module/item/items-collection.ts
```diff
@@ -1,11 +1,10 @@
 import { INameable } from '@module/data/mixins/nameable.js'
 
 class GurpsItems extends foundry.documents.collections.Items {
-  // @ts-expect-error - Something wrong with the return type?
   override getName<Options extends foundry.documents.abstract.DocumentCollection.GetOptions | undefined = undefined>(
     name: string,
     options?: Options
-  ): Collection.GetReturnType<Item.Stored, Options> {
+  ): Item.Stored {
     let entry = this.find(item => item.name === name)
```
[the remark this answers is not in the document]
mjeffw: I didn't make this change, so I'm not fixing it as I don't know the implications.

src/module/actor/actor-sheet.js
```diff
@@ -1156,77 +1155,23 @@ export class GurpsActorSheet extends foundry.appv1.sheets.ActorSheet {
...
 
-  async _editTracker(path) {
-    let tracker = foundry.utils.getProperty(this.actor.system, path)
-    let dialog = new ResourceTracker.TrackerEditor(JSON.parse(JSON.stringify(tracker)))
+    const path = pathAndKey.split('.').slice(0, -1).join('.')
+    const key = pathAndKey.split('.').slice(-1)[0]
 
-    dialog._updateTracker = async () => {
-      let update = {}
+    const trackers = foundry.utils.getProperty(this.actor, `system.${path}`)
```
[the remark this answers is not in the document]
mjeffw: No, `path` is set to `additionalresources.tracker` and does not start with `system`.

src/module/resource-tracker/resource-tracker.ts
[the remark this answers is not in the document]
mjeffw: Fixed by moving this logic to the Actor.

src/module/actor/modern/sheet.ts
```diff
@@ -761,4 +796,100 @@ export class GurpsActorModernSheet extends SheetBase {
...
+
+    if (!trackerKey) return
+
+    const trackers = (this.actor.system as Actor.SystemOfType<'character' | 'characterV2'>).additionalresources?.tracker
+    const tracker = trackers?.get(trackerKey)
+
+    if (!tracker) return
+
+    GURPS.modules.ResourceTracker.updateResourceTracker(this.actor, tracker)
+  }
```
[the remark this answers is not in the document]
rinickolous: I agree

src/module/actor/modern/sheet.ts
```diff
@@ -761,4 +801,102 @@ export class GurpsActorModernSheet extends SheetBase {
...
+
+    if (!trackers) return []
+
+    // Convert from Record<string, Tracker> to array of PreparedTrackerData for easier handling in templates.
+    const preparedData: PreparedTrackerData[] = []
+
+    let index = 0
+
+    for (const [key, tracker] of trackers.entries()) {
+      const trackerId = String(tracker.id ?? key)
```
rinickolous: Why is this needed if we're expecting `tracker.id` to always be defined?

src/module/actor/actor-sheet.js
```diff
@@ -1156,77 +1155,30 @@ export class GurpsActorSheet extends foundry.appv1.sheets.ActorSheet {
...
-
-    dialog.render(true)
-  }
+    const key = pathAndKey.split('.').slice(-1)[0]
 
-  async _editTracker(path) {
-    let tracker = foundry.utils.getProperty(this.actor.system, path)
-    let dialog = new ResourceTracker.TrackerEditor(JSON.parse(JSON.stringify(tracker)))
+    const trackers = this.actor.system.additionalresources.tracker || {}
+    const tracker = trackers[key]
```
rinickolous: I'm not sure whether this will reliably get the tracker entry, given that the queried object is a Collection.

src/module/actor/gurps-actor.ts
```diff
@@ -561,6 +560,44 @@ class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> impleme
...
+          }
+
+          const updates: Partial<TrackerInstance> = {}
+
+          updates.max = value
+          updates.value = tracker.isAccumulator ? tracker.min : value
+          updates.initialValue = null
+
+          tracker.update(updates)
+        }
```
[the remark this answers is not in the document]
rinickolous: I still agree with this. We should not be running any recursive update calls, and calling `tracker.update` runs an update call inside of data preparation, which will in turn run data preparation again.

src/module/actor/gurps-actor.ts
```diff
@@ -561,6 +560,44 @@ class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> impleme
...
+      }
+    } else {
+      // TODO Implement custom trackers or pools in GcsCharacter.
+      ui.notifications!.warn(
+        'GCS Characters do not currently support Resource Trackers. This may be added in a future update.'
+      )
+    }
```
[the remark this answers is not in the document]
rinickolous: The solution to this as far as I know is actually to not run this data preparation on the Document level at all. All type-specific data preparation should run only on the affected system model (so this would be moved to the `prepareDerivedData` method on `CharacterModel` instead of being on `GurpsActorV2`.

src/module/otf/description-utilities.ts
```diff
@@ -0,0 +1,226 @@
...
+    const resourceValue = foundry.utils.getProperty(actor, `system.${key}.value`) as number
+
+    if (resourceValue === undefined || resourceValue === null) {
+      ui.notifications!.warn(game.i18n!.format('GURPS.otf.resourceNotFound', { resource: key }))
+
+      return
+    }
+
+    await actor.update({ [`system.${key}.value`]: resourceValue - cost })
+  } else if (target.match(/^tr\d+/i)) {
```
rinickolous, line 63: Why is this attempting to match resource trackers using the legacy system? This would not have any use cases after automatic migration to `characterV2`.

src/module/otf/description-utilities.ts
```diff
@@ -0,0 +1,226 @@
...
+
+    const tracker = (actor.trackersByName as Record<string, any>)[trackerName]
+
+    if (!tracker) {
+      ui.notifications!.warn(game.i18n!.format('GURPS.otf.trackerNotFound', { tracker: trackerName }))
+
+      return
+    }
+
+    const trackerCurrentValue = foundry.utils.getProperty(
```
rinickolous: Why is this trying to access the tracker value using `getProperty`? The tracker object itself has already been obtained by the method, so why not just do `trackerCurrentValue = tracker.value`?

src/module/resource-tracker/ui/resource-tracker-editor-v2.ts
```diff
@@ -0,0 +1,346 @@
...
+    this.#tracker.pdf = ''
+    this.#tracker.alias = ''
+    this.#tracker.isDamageType = false
+    this.#tracker.isAccumulator = false
+    this.#tracker.isMaxEnforced = false
+    this.#tracker.isMinEnforced = false
+
+    this.#tracker.thresholds = []
+
+    await this.render({ force: true })
```
rinickolous, line 111: The `force` flag is unnecessary here (and in other instances of `this.render` for other application actions in this file). In all instances, the method is triggered by an interaction with the application, which necessitates that the application is already on-screen and therefore does not need to be force-rendered.

src/module/resource-tracker/ui/resource-tracker-editor-v2.ts
```diff
@@ -0,0 +1,346 @@
...
+        this.#setTrackerField(field, value)
+      })
+    })
+  }
+
+  #bindThresholdInputs(): void {
+    const root = this.element
+
+    root.querySelectorAll<HTMLInputElement | HTMLSelectElement>('[name][data]').forEach(element => {
+      element.addEventListener('change', event => {
```
rinickolous, line 244: This implementation may work just fine, but given that thresholds are complex objects, we may want to consider just making them into PseudoDocuments and storing them in a Collection on the Resource Tracker. This lets us avoid potential mishaps with updating arrays (not here, but elsewhere) and needing to deal with manual handling of updates which may affect an array.

src/module/resource-tracker/ui/resource-tracker-editor-v2.ts
```diff
@@ -0,0 +1,346 @@
...
+  /**
+   * Don't directly update the underlying tracker using `tracker.update({...})` when form fields change, as we don't
+   * want to mutate the tracker state directly. Instead, we update the tracker data in memory and only persist the changes
+   * when the user clicks the "Save" button. This allows the user to abort changes by clicking "Cancel" without affecting
+   * the original tracker data.
+   */
+  #bindFieldInputs(): void {
+    const root = this.element
+
+    root.querySelectorAll<HTMLInputElement | HTMLSelectElement>('[data-field]').forEach(element => {
```
rinickolous, line 208: This implementation may work just fine, but there might be a simpler way to do the same thing. DataModel instances can be cloned using the `.clone` instance method, creating an independent copy. Mutating this copy through the built-in update handling lets us targed more deeply stored keys (if any are added in the future) and preserves the other benefits of the DataModel (object validation and cleaning, if any are implemented, e.g. enforcing minimum/maximum values, etc.)

src/module/resource-tracker/ui/update-resource-tracker.ts
```diff
@@ -0,0 +1,18 @@
+import { TrackerInstance } from '../resource-tracker.js'
+import { IResourceTracker } from '../types.js'
+
+export async function updateResourceTracker(actor: Actor.Implementation, tracker: TrackerInstance): Promise<void> {
```
rinickolous, line 6: `actor` doesn't seem to be used in this function.

src/module/resource-tracker/ui/update-resource-tracker.ts
```diff
@@ -0,0 +1,18 @@
...
+  const trackerData = JSON.parse(JSON.stringify(tracker)) as IResourceTracker
```
rinickolous: I'm not sure why the Tracker instance is reduced to an object here and then updated indirectly. Why not just pass the instance through directly?

src/module/resource-tracker/ui/update-resource-tracker.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  interface TrackerEditorOptions {
+    onUpdate: (editedTracker: IResourceTracker) => Promise<void>
+  }
+
+  const dialog = new GURPS.modules.ResourceTracker.TrackerEditorV2(trackerData, {
```
rinickolous: `TrackerEditorV2` is being treated like a dialog here, but extends `ApplicationV2`. Why not make it extend `DialogV2` instead of using a workaround?

src/module/resource-tracker/ui/resource-tracker-editor-v2.ts
```diff
@@ -0,0 +1,346 @@
...
+  onUpdate?: (trackerData: IResourceTracker) => Promise<void> | void
+  onCancel?: () => void
+}
+
+type ResourceTrackerContext = foundry.applications.api.ApplicationV2.RenderContext & {
+  tracker: IResourceTracker
+}
+
+export class ResourceTrackerEditorV2 extends foundry.applications.api.HandlebarsApplicationMixin(
+  foundry.applications.api.ApplicationV2
```
rinickolous, line 18: Mentioned elsewhere too, but if this class is being used as a Dialog then it should extend DialogV2 instead.

src/module/resource-tracker/index.ts
```diff
@@ -110,10 +61,26 @@ async function convertOldSettings(
 interface ResourceTrackerModule extends GurpsModule {
   TemplateManager: typeof ResourceTrackerManager
   TrackerEditor: typeof ResourceTrackerEditor
+  TrackerEditorV2: typeof ResourceTrackerEditorV2
+  updateResourceTracker(actor: Actor.Implementation, tracker: TrackerInstance, trackerId?: string): Promise<void>
```
rinickolous: Explicitly defining a function signature here means needing to update it if the function signature ever changes. Consider just defining this as `typeof updateResourceTracker` (and as appropriate for other functions) to avoid this in the future. Additionally, this function signature does not match the one in the file which defined this function, which may lead to developer confusion when using the function if the developer checks only this signature and not the actual one.

src/module/resource-tracker/migration.ts
```diff
@@ -0,0 +1,60 @@
...
+  for (const oldTemplate of objectToArray(oldTemplates)) {
+    const newTemplate = new ResourceTrackerTemplate({
+      tracker: {
+        ...oldTemplate.tracker,
+        isAccumulator: oldTemplate.tracker.isDamageTracker ?? false,
+        useBreakpoints: oldTemplate.tracker.breakpoints ?? false,
+      },
+      initialValue: oldTemplate.initialValue,
+      autoapply: !!oldTemplate.slot,
+    })
```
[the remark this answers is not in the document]
rinickolous: Not true, `id` is auto-filled if not provided.

src/module/resource-tracker/migration.ts
```diff
@@ -0,0 +1,61 @@
...
+      delete (newTemplate.tracker as Record<string, unknown>).isDamageTracker
+    }
+
+    // remove old breakpoints field if it exists
+    if ('breakpoints' in newTemplate.tracker) {
+      delete (newTemplate.tracker as Record<string, unknown>).breakpoints
+    }
+
+    newTemplates[oldTemplate.tracker.name] = newTemplate
+  }
```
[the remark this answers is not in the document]
rinickolous: This is fine given order is preserved as far as I can tell.

src/module/resource-tracker/ui/resource-tracker-manager.ts
```diff
@@ -0,0 +1,281 @@
+import { TrackerComparators, TrackerOperators } from './types.js'
+import { IResourceTracker, IResourceTrackerTemplate, SETTING_TRACKER_TEMPLATES } from './types.js'
+import { ResourceTrackerEditorV2 } from './ui/resource-tracker-editor-v2.js'
+
+export class ResourceTrackerManager extends FormApplication {
```
rinickolous: It would be nice to minimally update this to `ApplicationV2` if it's being updated anyway.

src/module/resource-tracker/ui/resource-tracker-manager.ts
```diff
@@ -0,0 +1,281 @@
...
+      const editedTracker = await this.#editTrackerTemplate(key, template.tracker)
+
+      if (!editedTracker) return
+
+      template.tracker = editedTracker
+      template.name = editedTracker.name
+      this.render(true)
+    })
+
+    html.find('[name="autoapply"]').on('change', ev => {
```
rinickolous: Defining events such as this can be avoided if the templates object is passed through as a minimal DataSchema which can be updated by the usual DataModel methods.

src/module/resource-tracker/settings.ts
```diff
@@ -0,0 +1,25 @@
...
+    type: ResourceTrackerManager,
+    restricted: true,
+    icon: 'fa-solid fa-square-dashed-circle-plus',
+  })
+
+  game.settings.register(GURPS.SYSTEM_NAME, SETTING_TRACKER_TEMPLATES, {
+    name: game.i18n.localize('GURPS.resourceTracker.template.title'),
+    scope: 'world',
+    config: false,
+    type: Object as any,
```
rinickolous: This should be more specific (i.e. an ArrayField or TypedObjectField containing the templates, or else a DataModel containing such a field (if necessary))

src/module/util/hitpoints.js
```diff
@@ -60,12 +62,12 @@ const hpConditions = {
 const fpConditions = {
   NORMAL: {
     breakpoint: _ => Number.MAX_SAFE_INTEGER,
-    label: 'GURPS.normal',
+    label: 'GURPS.status.Healthy',
     style: 'normal',
   },
```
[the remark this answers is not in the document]
rinickolous: Given these are FP conditions, the status here should be `GURPS.status.Rested`.

src/rules/injury/fatigue.ts
```diff
@@ -0,0 +1,21 @@
...
+   * @returns An array of threshold descriptors, each containing a value and a condition.
+   */
+  static getThresholds(maxFP: number): ThresholdDescriptor[] {
+    if (!maxFP || maxFP <= 0) return []
+
+    return [
+      { value: maxFP, condition: 'GURPS.status.Rested' },
+      { value: Math.ceil(maxFP / 3) - 1, condition: 'GURPS.status.Tired' },
+      { value: 0, condition: 'GURPS.collapse' },
+      { value: -maxFP, condition: 'GURPS.unconscious' },
```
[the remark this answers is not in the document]
rinickolous: I think I agree with this.

src/rules/injury/hit-points.ts
```diff
@@ -0,0 +1,28 @@
...
+   *  and death.
+   * @returns An array of threshold descriptors, each containing a value and a condition.
+   */
+  static getThresholds(maxHP: number): ThresholdDescriptor[] {
+    if (!maxHP || maxHP <= 0) return []
+
+    return [
+      { value: maxHP, condition: 'GURPS.status.Healthy' },
+      { value: Math.ceil(maxHP / 3) - 1, condition: 'GURPS.status.Reeling' },
+      { value: 0, condition: 'GURPS.collapse' },
```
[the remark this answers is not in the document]
rinickolous: I think I agree with this.

static/lang/de.json
```diff
@@ -1380,7 +1406,7 @@
     "settings": {
       "physicalDiceTooltip": "z.B.: 1,2,3 oder 6"
     },
-    "settingSheetDetail": "Quick Sheet",
+    "settingSheetDetail": "Schnellbogen",
```
rinickolous: This is a bit of a literal translation from the German, literally a fast sheet of paper. I would maybe go for Wechselblatt (swap sheet / swappable sheet) to convey the meaning a bit better.

static/lang/ru.json
```diff
@@ -1495,7 +1521,7 @@
     "splitDrValue": "Значение СП",
     "splitQuantity": "Разделить количество",
     "spoken": "Устный",
-    "statsTab": "Stats",
+    "statsTab": "Характеристики",
```
rinickolous: I would probably go for `Статы` here, it's a Russian loanword from the English and is closer in length to the English, making layout a bit easier.

static/templates/resource-tracker/resource-tracker-editor.hbs
```diff
@@ -0,0 +1,135 @@
...
+    </button>
+    <button id='clear-button' type='button' class='ms-button ms-delete-button' data-action='clearTracker'
+      data-tooltip='{{localize "GURPS.resourceTracker.edit.clear"}}'>
+      <i class='fas fa-eraser'></i>
+    </button>
+  </header>
+
+  <div class='ms-content'>
+    <section class='text-attributes ms-section'>
+      <div class='ms-field'>
```
rinickolous, line 16: Nothing wrong with this really but we can cut down on explicitly defining the tags here (and potentially having to make repetitive changes later if we use the `formGroup` or `formInput` built in helpers). We'd have to pass through the actual class instance instead of a plain object for that, though I don't see any reason not to.

src/module/actor/gurps-actor.ts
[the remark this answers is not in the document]
mjeffw: I still agree with you, I still don't know how to fix it. Yes, the data prep will run again, but I don't see any other solution.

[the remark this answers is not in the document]
rinickolous: I'll try to take a look later today to see if there's another way around this. Bear with.

src/module/actor/gurps-actor.ts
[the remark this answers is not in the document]
mjeffw: I don't see how moving this to the data model solves anything. We will still need to tell the user that the tracker cannot be updated. The only solution I can offer is that we always set the `initialValue` field to null after trying to resolve the value whether it works or not. That means the data prep will run only twice, and only once will the user be told there is a problem.

src/module/actor/gurps-actor.ts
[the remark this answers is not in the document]
rinickolous: Looking at this again, I'm not quite sure what this data preparation method is trying to achieve. Could you explain it to me?

src/rules/injury/hit-points.ts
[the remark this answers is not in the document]
mjeffw: The rule is the character is Reeling if there is "Less than 1/3 your HP left" -- Not exactly 1/3, but less than 1/3. Thus someone with 15 HP is NOT reeling at current HPs = 5; but is reeling at HPs = 4.

src/module/util/hitpoints.js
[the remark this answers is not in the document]
mjeffw: ?? It is in my view of the code:

```
      { value: maxFP, condition: 'GURPS.status.Rested' },
```

src/rules/injury/fatigue.ts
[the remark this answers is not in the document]
mjeffw: The algorithm is correct.

src/module/actor/modern/sheet.ts
mjeffw: Fixed

src/module/actor/actor-sheet.js
mjeffw: Fixed

static/lang/de.json
mjeffw: Fixed.

static/templates/resource-tracker/resource-tracker-editor.hbs
mjeffw: What's your suggestion?

src/module/importer/migrate.ts
rinickolous, line 1: very clean, big fan of this

src/module/util/constrast-color.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Determine the luminescence of the provided background color, and return either a light or dark foreground color to
+ * contrast with it.
+ *
+ * @param {string} backgroundHex - The background color as a 6-digit hex string (e.g., '#ff0000' for red).
+ * @param {string} darkColor - The color to use for dark foreground elements.
+ * @param {string} lightColor - The color to use for light foreground elements.
+ * @returns {string} - The appropriate foreground color based on the background color's luminescence.
+ */
+export function constrastColor(backgroundHex: string, darkColor = '#1c1a17', lightColor = '#f8f6f2'): string {
```
rinickolous: typo in function name as copilot said

src/module/util/contrast-color.ts
rinickolous, line 1: typo in filename

src/module/resource-tracker/ui/resource-tracker-editor-v2.ts
mjeffw: The advice in the Foundry Community Wiki is to use ApplicationV2 if the window needs to be re-rendered. That will happen here if the user clears the tracker, or clones a template.

src/module/actor/modern/dialog-crud-handler.ts
```diff
@@ -167,12 +168,216 @@ export function bindNoteCrudActions(
...
+        `${path}.${trackerKey}`,
+        tracker.name,
+        game.i18n!.format('GURPS.resourceTracker.trackerNamed', { name: tracker.name }),
+        async _ => {
+          await tracker.delete()
+        }
+      )
+
+      if (confirmed) {
+        await actor.refreshDR()
```
rinickolous, line 232: Not specifically a comment about this line but it seems to me like the underlying `drMod` and other properties changed by the `refreshDR` method may be better suited as derived properties rather than the persistent properties they currently are, and that rather than calling this method manually, the properties should refresh automatically during data preparation.

That's not a change request for this PR, I'd just like to hear your thoughts on the matter.

src/module/actor/modern/sheet.ts
```diff
@@ -82,6 +99,11 @@ export class GurpsActorModernSheet extends SheetBase {
       importActor: GurpsActorModernSheet.#onImportActor,
       editQuickNotes: GurpsActorModernSheet.#onEditQuickNotes,
       editMoveMode: GurpsActorModernSheet.#onEditMoveMode,
+      // editTracker: GurpsActorModernSheet.#onEditTracker,
```
rinickolous: just needs removing

src/module/util/contrast-color.ts
```diff
@@ -0,0 +1,32 @@
+'use strict'
```
rinickolous: don't need `use strict` in a TS file

src/module/resource-tracker/ui/resource-tracker-manager-v2.ts
```diff
@@ -0,0 +1,327 @@
+import { DeepPartial } from 'fvtt-types/utils'
+
+import { TrackerComparators, TrackerOperators } from '../types.ts'
+import { IResourceTracker, IResourceTrackerTemplate, SETTING_TRACKER_TEMPLATES } from '../types.ts'
+
+import { ResourceTrackerEditorV2 } from './resource-tracker-editor-v2.ts'
```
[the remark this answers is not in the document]
mjeffw: Fixed

src/module/actor/modern/dialog-crud-handler.ts
mjeffw: Agreed!

src/module/actor/modern/sheet.ts
mjeffw: done

src/module/util/contrast-color.ts
mjeffw: Fixed

PR #2580, opened 2026-03-14
Feature: GCS Actor Sheet V2

rinickolous:
This PR adds an ApplicationV2 actor sheet meant to mimic the original GCS sheet as closely as practicable.

The sheet is exclusive to Actors of type `characterV2` and uses `characterV2`-specific keys in some places. Some effort has been made to make sure actors of other types (thinking of `gcsCharacter` in particular) can pass their data and schema through an adapter to display correctly on the sheet. However, this doesn't apply to the whole sheet as of yet (improvements will be made in the future).

A notable feature in this sheet is the `PoolValueElement` used for pool displays. This element allows users to input a pool value directly and update the new `damage` value for pools (mentioned below) without requiring custom processing of the update data.

This PR also expands slightly on @daynin 's base actor sheet, making this its own sheet with some shared options and methods attached. I believe basing any new custom actor sheets on this base sheet would be a good idea going forward.

I borrowed some files from @mjeffw 's in progress pool values / resource tracker's PR (the re-done HP and FP threshold functions specifically) in anticipation of those PRs being merged anyway.

The `types/` files in this PR include what basically amounts to namespace shortcuts used to maintain stability of type references in cases where `fvtt-types` changes theirs.

---

This PR includes some changes to data schemas, as follows:
- Character Pools (HP, FP, QP) on `characterV2` now use a `damage` field instead of a `value` field, which makes enforcement of maximum values much easier. The `value` property is now an accessor and functions as before.
- The Move Mode schema no longer includes a `default` checkbox, with the `characterV2` schema instead containing a `_currentMoveModeId` field indicating the currently selected move mode. I changed this to make managing the move mode easier algorithmically, as the previous method required toggling the `default` boolean on other move modes if a given move mode is set as the default.
- `equipmentV2`'s `carried` field is not `_carried`, with `carried` being moved to an accessor. Said accessor will return the value of `_carried` for top-level items, but will defer to the `carried` accessor of the containing item otherwise. I changed this to make managing carried / uncarried items easier without needing to update the data of child items if their parent item is moved from carried to uncarried or vice-versa. This effectively eliminated the possibility of a child item being treated as carried (for weight / weapon enablement purposes) when its parent is not carried.
- Added a `techlevel` field to `spellV2`, because Spells can have tech levels.

This PR includes some additions to various models, generally for the sake of being able to easily adapt different types of models into objects which the sheet can interact with easily. These are as follows:
- added `toggleOpen()` as a method to the `IContainable` mixin. Implementations should use their own logic for performing the action (`BaseItemModel` already implements this)
- Added a `sortKeys` property to `ItemMetadata`. This is a `Record<string, string>` matching a key to a system path, allowing items in a list to be sorted according to a given property value. The V2 item models implement this. Setting it as a static property tied to the item type will allow sheets to support both V2 item models and future GCS ones by eliminating reliance on a specific system path.
- Added a `toDisplayItem()` method to all V2 items. This returns a plain display object with a specific known schema, used for displaying on the character sheet. This, again, is meant for future adaptations when GCS model support is added to the sheet. The display types are defined independently of the schemas, meaning that if any changes are made to them, the LSP should prompt the developer to update the corresponding methods in all affected model classes.

This PR also includes some bug fixes for issues I found while testing the added features:
- Move modes were duplicated on import. Fixed this by instead modifying move modes of the same name as imported move modes on import.
-

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,6 +20,7 @@
         "prettier": "^3.5.3",
         "purgecss": "^8.0.0",
         "sass": "^1.97.1",
+        "showdown": "^2.1.0",
         "tsc-alias": "^1.8.16",
         "tsc-watch": "^7.2.0",
         "typescript": "^5.8.3",
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "prettier": "^3.5.3",
     "purgecss": "^8.0.0",
     "sass": "^1.97.1",
+    "showdown": "^2.1.0",
     "tsc-alias": "^1.8.16",
     "tsc-watch": "^7.2.0",
     "typescript": "^5.8.3",
--- a/test/vitest.setup.ts
+++ b/test/vitest.setup.ts
@@ -1,3 +1,5 @@
+import showdown from 'showdown'
+
 export {}
 
 const typePrototypes: [new (...args: never[]) => unknown, string][] = [
@@ -239,6 +241,8 @@ class MockCollection<V> extends Map<string, V> {
     return result
   }
 }
+
+globalThis.showdown = showdown
 
 global.foundry = {
   abstract: {


src/module/actor/sheets/gcs-sheet.ts
```diff
@@ -0,0 +1,649 @@
...
+    }
+
+    const maxPath = systemPath.replace(/\.value$/, '.max')
+    const maxValue = foundry.utils.getProperty(this.actor, maxPath)
+    const newValue =
+      valueDelta > 0 && typeof maxValue === 'number'
+        ? Math.min(pathValue - valueDelta, maxValue)
+        : pathValue - valueDelta
+
+    await this.actor.update({ [systemPath]: newValue })
```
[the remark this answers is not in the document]
rinickolous: This is actually fine, it only applies to HP, FP and QP pools for now, and handling other resource trackers will be implemented later.

src/module/actor/data/move-mode.ts
```diff
@@ -22,10 +22,12 @@ class MoveModeV2 extends PseudoDocument<MoveSchema> {
 const moveSchema = () => {
   return {
     ...pseudoDocumentSchema(),
+    /* The name of this move mode */
     mode: new fields.StringField({ required: true, nullable: false }),
+    /* The "basic" value of this move mode, this is the normal move value for the mode. */
     basic: new fields.NumberField({ required: true, nullable: false, initial: 0 }),
+    /* The enhanced move value for this move mode.*/
     enhanced: new fields.NumberField({ required: true, nullable: true }),
-    default: new fields.BooleanField({ required: true, nullable: false, initial: false }),
```
mjeffw, line 28: What's the plan for displaying the default move mode? My intention was to set the move of the character to the default mode when opening the character sheet. For example, an avian species would have a 'Ground' move, but the default should be 'Air'.

src/module/actor/modern/sheet.ts
```diff
@@ -1,45 +1,24 @@
...
-  ActorSheetV2Configuration,
-  ActorSheetV2RenderOptions,
-  ActorSheetV2RenderContext,
-  HandlebarsTemplatePart,
-  HeaderControlsEntry,
-  HandlebarsActorSheetV2Constructor,
-} from '@gurps-types/foundry/actor-sheet-v2.js'
 import { Application } from '@gurps-types/foundry/application.js'
-import { getGame, getUser } from '@module/util/guards.js'
+import { getGame } from '@module/util/guards.js'
```
mjeffw, line 2: I really need to get in the habit of using this method to access the global game.

src/module/actor/sheets/gcs-sheet.ts
```diff
@@ -0,0 +1,662 @@
...
+  }
+
+  /* ---------------------------------------- */
+  /*  Drag & Drop Handling                    */
+  /* ---------------------------------------- */
+
+  protected override _onDragStart(event: DragEvent) {
+    const element = event.currentTarget
+
+    if (!isHTMLElement(element)) return
```
mjeffw: I have code like this -- but when I make a mistake or a bad assumption, it returns without an error. I wonder if we should instead throw an exception or at least write something to the console?

src/rules/injury/conditional-injury/conditional-injury.js
mjeffw, line 1: Just an FYI -- in the future, I want to create the standard GURPS damage and injury system as a generic interface that injects code into everywhere damage is rolled or injury calculated, and make Conditional Injury a drop-in replacement.

static/templates/actor/gcs/equipment.hbs
```diff
@@ -0,0 +1,174 @@
+{{#*inline ".equipmentEntry"}}
```
mjeffw, line 1: Just for my information, what does this helper do?

mjeffw (approved):
I left a few comments and questions, but nothing that would hold up merging and closing this PR. I'll leave it to you to address those or not and merge/close.

static/templates/actor/gcs/equipment.hbs
rinickolous: it's an inline partial. We actually have lots of potential use-cases for these. Turns out you don't need a whole file every time you want to iterate something, you can just define the partial in the parent file itself.

src/module/actor/data/move-mode.ts
rinickolous: I moved it over to a persistent property on the character itself, so once it's set for a given character it'll stay that way. I hadn't really considered that we might want to set the default to something other than ground but I'll just make sure the default stays persistent between imports and I think that should be enough.

src/module/actor/sheets/gcs-sheet.ts
rinickolous: For sure, I'll change that now.

src/rules/injury/conditional-injury/conditional-injury.js
rinickolous: Sounds great.

PR #2617, opened 2026-04-04
Feature/refactor gcs actor css

### LINKED ISSUE #2606: V1.0.0 Fix space issue with GCS Actor sheet
There's inconsistent spacing between Primary and Secondary attributes and Basic Damage. I prefer the smaller spacing of the Primary attributes.

The spaces indicated by the red markup should be exactly the same:

Image

### LINKED ISSUE #2607: V1.0.0 Actor sheet tooltip styling
>> For tooltips, lets use Foundry's built-in data-tooltip or data-tooltip-html. (Referencing the tooltips on the trackers.

> That's what I'm using. I just changed the styling to resemble the GCS style.

I strongly suggest either: a) Just use the default tooltip appearance with Foundry -- this keeps our tooltips and Foundry's consistent, or b) define our own unique and globally applied styling.

I prefer (a) since we don't have to do anything to support it.

### LINKED ISSUE #2615: V1.0.0 HP and FP thresholds
The `@rules` source code module should not contain any UI-only data. The 'color' attribute of the thresholds violate this principle. Condition is fine -- wording should align with rules, but there is no standard rules-based color values.

There should be another place that defines the colors -- perhaps a world setting? -- and the threshold data should be enhanced with that value.

This global color definition may be defined uniquely or per-sheet.

In any case, we should not be hard-coding color values in the source code in a way that the user can't override. 

Also applies to Conditional Injury.

**Bug:** Color declaration in the rules/injury code.
**Feature:** Make color configurable.

### LINKED ISSUE #2620: V1.0.0 GCS Actor Sheet Secondary attributes vertical alignment
Image

Make the right edge of the 6's line up with the right edge of the 15s.

mjeffw:


src/styles/_variables.scss
```diff
@@ -7,3 +7,70 @@ body.theme-dark {
...
+  --gga-color-med-orange: #c38355;
+  --gga-color-dark-orange: #794b2a;
+
+  --gga-color-red: #e45144;
+  --gga-color-med-red: #9c423a;
+  --gga-color-dark-red: #4e211d;
+
+  /* ---------------------------------------- */
+
+  --gga-color-modifier-0: var(--gga-color-green);
```
rinickolous, line 51: This is a lot nicer than my idea, I like it

rinickolous (commented):
The input fields could use a bit of a border so they don't blend together in certain situations like the description block here. Otherwise, I like the change to a more sepia tone and the muted colors in dark mode and welcome these changes.

Image

PR #2635, opened 2026-04-09
PseudoDocument Refactor

rinickolous:
This PR depends on #2634 . This PR:
- Adds the `pseudoCollections` object to Actors and Items, used to easily access collections of PseudoDocuments by document name, regardless of the system path under which the location resides.
- Enhanced the `getEmbeddedDocument`, `createEmbeddedDocument`, `deleteEmbeddedDocument`, and `getEmbeddedCollection` methods for Actors and Items to better support PseudoDocument collections.
- Changes the data preparation and initialization process for ModelCollection and CollectionField. The new process should be more straightforward.
- Adds an option to choose to remove (or not remove) the contents of a PseudoDocument when it is deleted.
- Changes the GCS sheet markup to work the new embedded document targeting actions.
- Adds default values for Actor hit locations and move types on creation.
- Adds the `embeddedHolderField` field to Actor metadata, used to specify the system path under which  an embedded document holder Item is stored. This is used for more easily interacting with pseudo-documents which are held by the holder item but visually appear to be held by the actor directly.
- Adds the `name`, `sort` keys to PseudoDocuments and removes duplicate fields in subclasses of PseudoDocument wherever they appear.
- Adds better mocks for Foundry util functions to Vitest.


src/module/config/hit-locations.ts
```diff
@@ -0,0 +1,213 @@
...
+  Partial<DataModel.CreateData<DataModel.SchemaOf<HitLocationEntryV2>>>
+> => {
+  const locations: Partial<DataModel.CreateData<DataModel.SchemaOf<HitLocationEntryV2>>>[] = [
+    {
+      name: '',
+      where: 'Eyes',
+      import: 0,
+      penalty: -9,
+      rollText: '-',
+      role: 'eye',
```
mjeffw: `role` is another place we should use a named constant, and that constant should be defined in the `@rules` module.

mjeffw: Actually, maybe, the default Hit Locations in general should be in `@rules`, except for fields used for internal stuff like `sort`.

test/foundry-utils/object-utils.ts
```diff
@@ -0,0 +1,289 @@
...
+  [Error, 'Error'],
+]
+
+/* ---------------------------------------- */
+
+const applySpecialKeys = (val: unknown): unknown => val
+
+/* ---------------------------------------- */
+
+export function getType(variable: unknown): string {
```
mjeffw, line 17: I'm (somewhat) okay with this (i.e., defining the function by copying the foundry code, more or less) but in general mocking functions for testing doesn't provide an implementation. Rather, the mocks have an interface that says "if I am given these parameters, I should return this value" and that's entirely intentional. We don't need to know how they work we only need to know the side effects of calling a mocked function.

mjeffw (changes requested):
I'm basically okay with the PR being merged but just wanted you to read and consider the few comments I made before doing so.

src/module/config/hit-locations.ts
rinickolous: Make sense. I've extracted the rules-only parts of the Hit Location to a `@rules` module and refactored this to generate a hit location table from said file, inserting the Foundry-specific fields in.

test/foundry-utils/object-utils.ts
rinickolous: Sure, I'll consider simplifying the functionality. I added this in because we were having some issues with the tests not matching Foundry behavior and the easiest resolution in mind at the time was just copying the functions at fault over.

PR #2636, opened 2026-04-09
Migration Refactor

rinickolous:
This PR does the following:
- Changes Actor and Item migration so that it occurs earlier, during DataModel initialization rather than after the `ready` hook.
- Removes the old Item and Actor types (enabled by the previous change, as the new migration migrates Documents without changing their type)
- Adds a `Migration` module to facilitate migration system-wide.
- Removes unused v1 Actor and Item files and methods.


src/module/actor/migrate.ts
```diff
@@ -1,548 +1,28 @@
...
-
-      newData.additionalresources ||= {}
-      newData.additionalresources.tracker ||= {}
-      newData.additionalresources.tracker[id] = trackerData
-    })
-  }
-
-  return newData
-}
+  if (systemVersion === null) return source
```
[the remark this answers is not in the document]
rinickolous: This wouldn't work as intended. Actor data is sometimes migrated before it is fully initialized (i.e. _stats field is not yet present). In such cases, we should not attempt migration.

src/module/actor/migrations/v1_0_0.ts
mjeffw, line 1: I'm not particularly happy with file names starting with a number.

src/module/actor/gurps-actor.ts
```diff
@@ -1986,17 +1994,19 @@ class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> impleme
    *
    * @deprecated Use moveItem() instead
    */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   async reorderItem(sourceKey: string, targetkey: string, object: any, isSrcFirst: boolean) {
     if (this.isNewActorType) return await this.moveItem(sourceKey, targetkey)
 
+    // NOTE: Deprecated. Method should never run past this point.
```
mjeffw: Throw an exception -- We would like to know when and how it somehow executed without `this.isNewActorType === true`.

Alternatively, if this code cannot be reached, delete it.

src/module/actor/gurps-actor.ts
```diff
@@ -2820,150 +2828,151 @@ class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> impleme
...
-    }
-
-    const data = arrayToObject(currentTrackers)
-
-    // Remove all trackers first. Add the new "array" of trackers.
-    if (data) {
-      await this.update({ 'system.additionalresources.-=tracker': null } as Actor.UpdateData)
-      await this.update({ 'system.additionalresources.tracker': data } as Actor.UpdateData)
-    }
+    // NOTE: Deprecated. Method should never run past this point.
```
mjeffw: Throw an exception.

src/module/actor/gurps-actor.ts
```diff
@@ -3370,173 +3379,173 @@ class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> impleme
...
-                  if (!!hitLocation.drCap && hitLocation.dr > hitLocation.drCap) hitLocation.dr = hitLocation.drCap
-                }
-                //console.warn(e.where, e.dr)
-              }
-            })
-          } // end DR
-        }
-      }
-    }
+    // NOTE: Deprecated. This method should never run past this point. In Actor v2, bonuses are applied automatically
```
mjeffw: Exception

src/module/actor/gurps-actor.ts
```diff
@@ -3554,150 +3563,154 @@ class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> impleme
...
-
-      return
-    }
+    if (this.isNewActorType) return
 
-    // New item created in Foundry v12.331 dragData:
-    // {
-    //   "type": "Item",
-    //   "uuid": "Item.542YuRvzxVx83kL1"
+    // NOTE: Deprecated. This method should never run past this point.
```
mjeffw: Exception

src/module/actor/gurps-actor.ts
```diff
@@ -4439,96 +4452,95 @@ class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> impleme
...
-        itemModifiers: item.modelV1.itemModifiers,
-      },
-    })
-    await this._addItemAdditions(item, sysKey)
-
-    if (item.type === ItemType.LegacyEquipment) {
-      await this.updateParentOf(sysKey, true)
-      await this._updateEquipmentCalc(sysKey)
-    }
+    // NOTE: Deprecated. Method should never run past this point.
```
mjeffw: Exception

src/module/actor/gurps-actor.ts
```diff
@@ -4673,105 +4685,112 @@ class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> impleme
...
-
-    await this.internalUpdate({ [nullKey]: null })
-
-    // Insert the element into the array.
-    array.splice(index, 0, element)
-
-    // Convert back to an object
-    object = arrayToObject(array, 5)
+  async _removeKey(_sourceKey: string) {
+    if (this.isNewActorType) return
```
mjeffw: Don't check for new type just throw an exception.

src/module/actor/gurps-actor.ts
```diff
@@ -4673,105 +4685,112 @@ class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> impleme
...
-      // simple eqt
-      const neqt = foundry.utils.duplicate(srceqt)
-
-      neqt.count = count
-      this.ignoreRender = true
-      await this.updateEqtCount(srckey, srceqt.count - count)
-      await GURPS.insertBeforeKey(this, targetkey, neqt)
-      await this.updateParentOf(targetkey, true)
-      this._forceRender()
+    if (this.isNewActorType) return
```
mjeffw: same

src/module/actor/index.ts
```diff
@@ -107,7 +45,7 @@ export const Actor: ActorModule = {
   init,
   dataModels,
   sheets,
-  migrate: runMigration,
+  migrations,
```
mjeffw, line 48: I would prefer that we keep a single `migration()` function in the Module interface. If there are multiple migrations, this function will call them making the decision on what order, whether to call it with `await` or in parallel, etc. There's no need to expose that level of detail outside the module interface. (Just think of the `migrate` function as "migrate the module" not a specific list of migrations.)

src/module/item/migrations/v1_0_0.ts
mjeffw, line 1: no numbers

src/module/migration/migrations/v1_0_0.ts
mjeffw, line 1: no file names starting with a number

mjeffw (changes requested):

mjeffw:
Make sure you resolve any merge conflicts.

src/module/actor/index.ts
rinickolous: I understand where you're coming from, but I think the new approach is cleaner and grants us some additional benefits:
- The central migration module's own migrations run before any other modules, so checks and emergency aborts (like the one implemented for the 1.0.0 migration) can be easily slotted in before any other migration is run.
- Modules do not have to implement their own version filtering logic. In the original approach this was just a version comparison, but it would be repeated across every module that defines migrations.

I understand that the `migrations` array exposes a bit more of a module's internals, but the tradeoff is worth it. The coordination benefits come precisely from the Migrator being able to see and sequence those entries.

If you think the above are not sufficient reasons to change the approach then I can switch to something closer to the old approach, but I really do think this makes implementing new migrations a lot easier and makes running said migrations more predictable.

PR #2639, opened 2026-04-10
Application Types & GCS Actor Sheet Item drop refactor

rinickolous:
This PR depends on #2636 . It:
- Reverts our custom type definitions for ActorSheet, ItemSheet, Application, and other Foundry classes and namespaces to the `fvtt-types` definitions for coverage completeness. The excessive type depth issues which previously necessitated custom type definitions have now been resolved.
- Refactors `_onDropItem` in `GurpsActorGcsSheet`, adds a notification on transfer, improves legibility of code, adds slider input for item quantity.
- Refactors `#onSortItems` in `GurpsActorGcsSheet` so Items are sorted hierarchically (for alphabetical sorting, item children are sorted in relation to their siblings).


src/module/actor/sheets/gcs-actor-sheet.ts
```diff
@@ -987,78 +981,213 @@ class GurpsActorGcsSheet extends GurpsBaseActorSheet<
...
+          // Delete the item and all its descendants — children are not automatically cascade-deleted
+          // when their container is removed, so we must include them explicitly.
+          const allSourceIds = [item.id!, ...item.system.allContents.map(descendant => descendant.id!)]
+
+          details.deletions.push({ ids: allSourceIds, operation: { parent: item.parent! } })
+        } else {
+          details.updates.push({
+            data: [{ _id: item._id, 'system.count': remainingQuantity } as Item.UpdateData],
+            operation: { parent: item.parent! },
+          })
```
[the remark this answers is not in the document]
rinickolous: Not an issue in practice because dragStart requires ownership but good fallback in case that changes so worth implementing.

src/module/importer/importer-prompt.ts
mjeffw: Make sure you don't accidentally undo the use of constants (`ItemType.Skill`, e.g.)

src/types/gurps-module.ts
mjeffw, line 5: As previously mentioned, having a method here instead of a list of migrations both hide the complexities of the migrations from the caller and gives more latitude for the modules to customize how they run the migrations.

mjeffw (changes requested):
Make sure you address merge conflicts.

mjeffw (approved):

PR #2640, opened 2026-04-11
Miscellaneous 1.0.0 bugfixes

### LINKED ISSUE #2610: V1.0.0 GCS Actor sheet text alignment issues
- Right justify the following elements:
  - Any modifier
  - Any skill level
  - Hit Location DR
  - Encumbrance Max Load, Move, and Dodge (alternatively, Move and Dodge could be centered in the same way their Headings are.)
  - Points section.
  - the Edit icon on the Quick Notes section:
    
Image

  - The Open/Close icons on weapons -- they need to be right justified, and the size and color modified to match those on Traits.

Image

- Center:
  - Resource Tracker value (this should be resolved by using the standard Trackers)

### LINKED ISSUE #2612: V1.0.0 GCS Actor sheet text justification
Ensure the vertical positioning of the label and value align:

Image

Combat Move, Actions per Turn, and Blocks per Turn labels are below the values. My preference is that the values be moved down vertically to align with the labels.

### LINKED ISSUE #2618: V1.0.0 GCS Actor Sheet vertical spacing
The vertical spacing of the fields in the IDENTITY and MISCELLANEOUS sections is greater than other sections. Compare to the POINTS section immediately to the side of these sections; see how the two fields in each of these sections takes up more room that two fields in the POINTS section.

### LINKED ISSUE #2619: V1.0.0 GCS Actor sheet horizontal alignment in the MANEUVER section
Add some padding to the non-UI control fields to align the text with those controls. (The left edge of the word "Full" should align with the left edge of "Standing".)

Image

### LINKED ISSUE #2621: V1.0.0 GCS Actor Resource Tracker editing doesn't work
As the title says.

### LINKED ISSUE #2624: V1.0.0 Import always asks to overwrite body plan even if it is the same

### LINKED ISSUE #2625: V1.0.0 GCS Import does not provide DR
Character in Foundry:

Image

Same character in GCS:

Image

rinickolous:
This PR depends on #2639. It consists of bugfixes and very small changes relating to the GCS and modern actor sheets and import functionality. Some of the issues were previously pointed out and are labeled as such, while other issues were introduced in the previous set of PRs. List of changes below:

- Changes label / input pairs on the GCS sheet where the label is on the left so that the label text is right-aligned instead of left.
- Adds the modern sheet and modern NPC sheet as options for the `character` model (they were previously removed by accident)
- Fixes resource tracker elements on the modern sheet referring to the Resource Tracker threshold state by the variable name `condition`
- Right-aligns numeric values and center aligns clickable values (fixes #2610 )
- Aligns the notes styling for attacks with the same styling for item notes on the GCS sheet
- Fixes the notes toggle buttons not working on the GCS sheet
- Vertically centers the input and span elements in the GCS sheet maneuver section (fixes #2612 )
- Adds left-padding to span elements in the GCS sheet maneuver section so they align with the text in accompanying select boxes (fixes #2619 )
- Fixes import logic issues caused by the GCA imported actor schema preferring blank string values for nullable strings even when `null` is specified as the initial value.
- Fixes an issue in which the hit location table would be duplicated on first import.
- Fixes an issue in which the user would be prompted for hit location table overwrite even when the tables are identical (fixes #2624 )
- Fixes RoF and Bulk not being imported for ranged attacks from GCA
- Fixes the prepareDerivedData method not running for pseudo-documents owned by an Item.
- Fixes DR values not being properly imported from GCS and GCA (fixes #2625 )
- Fixes an issue in which editing the current value of a resource tracker on the GCS sheet doesn't work. (fixes #2621 )
- Fixes an issue in which pressing the Enter key on the GCS sheet would cause the HP value to decrement.
- Fixes Legality Class not being imported for equipment from GCA.
- Adds the `organization` profile field from GCS and displays it and the `player` field in the GCS sheet header section.
- Ensures that the portrait is always square on the GCS sheet (related to and potentially closes #2618 )
- Adds back accidentally removed status localization entries for English (affects HP, FP, and CI thresholds)


src/styles/gcs-sheet/_base.scss
```diff
@@ -323,6 +333,10 @@ body {
       background-color: var(--gcs-color-surface-bg);
     }
 
+    td + td {
```
mjeffw, line 336: I made some changes to the GCS sheet including I believe removing the border between columns in some tables, which I believe this CSS rule to be putting back. Let's talk if you don't agree with my decisions, I just don't want us to flip-flop without discussion.

static/templates/actor/gcs/resource-trackers.hbs
```diff
@@ -20,7 +20,7 @@
   </div>
   <div class="gcs-section-content">
     <div class="form-fields flexrow">
-      <button data-action="decrementPool" data-inverted="{{invertedDelta}}" data-path="{{numerator.path}}" {{#if atMin}}disabled{{/if}}><i class="fa-solid fa-minus"></i></button>
+      <button type="button" data-action="decrementPool" data-inverted="{{invertedDelta}}" data-path="{{numerator.path}}" {{#if atMin}}disabled{{/if}}><i class="fa-solid fa-minus"></i></button>
```
mjeffw, line 23: This might be another place I made changes -- I changed the appearance of the +/-/reset buttons to just look like the icons with no border or background.

mjeffw (changes requested):

src/styles/gcs-sheet/_base.scss
rinickolous: Sure. I think adding the column border here improves legibility in cases where the value in the column is close to the edge. Why not keep it?

static/templates/actor/gcs/resource-trackers.hbs
rinickolous: This change doesn't affect that, it just fixes an issue where if you don't specify `type="button"` for a button then it assumes it's a submit button and imitates pressing it when you press Enter anywhere on the sheet.

However, since you bring it up, I think your styling of those buttons is a bit hard to read. I just made some adjustments. Compare this:
image
image
to this:
image
image

I believe mine more clearly read as buttons, have more room to breathe (I added some gaps to the whole tracker element, it was rather cramped previously), and more clearly advertise state (disabled vs. enabled) than the plain buttons.

src/styles/gcs-sheet/_base.scss
mjeffw: I'll accept it right now, but reserve the right to remove it later. In general the UI should minimize (IMO) the number of borders, dividing lines, etc., all of which are "visual noise" that detracts from the clean look.

static/templates/actor/gcs/resource-trackers.hbs
mjeffw: Mine certainly needed the gaps, which is a separate issue. I do think your new approach is better than your old one. Again, I'll reserve judgment until I have a chance to pull up the UI and review it in situ.

mjeffw (approved):

PR #2647, opened 2026-04-22
Bugfix: Various bugfixes for FoundryVTT v14 Compatibility

rinickolous:
This PR fixes the following issues:
- Actor of type `character` fails to initialize due to missing Basic Move value
- Item of type `feature` causes migration to fail if `cr` value is `undefined`.
- Old style `-=` field deletions are incompatible with FoundryVTT v14 (replaced these).
- Cannot import Actors due to missing fields for Pseudo-Documents in the import data.

This PR also bumps the minimum FoundryVTT version to 14. Only the field deletions are a true breaking change here so we could apply the same compatibility patch that currently exists for the `0.18.16` version, but I think just sticking to v14 for this version will be cleaner.


src/module/actor/gurps-actor.ts
```diff
@@ -1022,35 +1022,10 @@ class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> {
...
-          [path + '.-=contains']: null,
-          [path + '.contains']: {},
-          [path + '.collapsed']: temp,
-        }
-
-        await this.update(update)
-      }
-    }
+    // Check if object implements IContainable interface and call toggleOpen
+    if (ContainerUtils.isToggleable(obj)) await obj.toggleOpen(expandOnly)
```
mjeffw, line 1028: Note to come back to this line.

src/module/importer/gcs-importer/importer.ts
```diff
@@ -571,7 +577,7 @@ Portrait will not be imported.`
     // On first import, always replace the hit location table
     if (this.actor && !this.actor.system.profile.modifiedon && !this.actor.system.additionalresources.importname) {
       const currentHitLocationNullifiers = Object.fromEntries(
-        this.actor.system.hitlocationsV2.map(location => [`-=${location._id}`, null])
+        this.actor.system.hitlocationsV2.map(location => [location._id, globalThis._del])
```
mjeffw, line 580: Is this the real pattern in v14?

mjeffw (approved):

PR #2667, opened 2026-04-27
Refactor: Modern Actor Sheet

rinickolous:
This PR removes outdated (0.18) functionality from the Modern Actor sheet and replaces it with equivalent functionality designed for use with the 1.0.0 data models. It moves around some methods which were previously specific to the GCS Actor Sheet (primarily the item drop handling) so that it can also be used by the Modern Actor Sheet.


src/module/actor/sheets/helpers.ts
```diff
@@ -95,6 +99,254 @@ export async function resolveItemDropQuantity(item: Item.OfType<ItemType.Equipme
 
 /* ---------------------------------------- */
 
+export async function resolveItemDropDetails(
```
mjeffw: This is a super long function; perhaps some of the details could be captured in other functions and the names of those functions aimed at improving readability of the code?

mjeffw (changes requested):

src/module/actor/sheets/helpers.ts
rinickolous: Done

mjeffw (approved):

PR #2695, opened 2026-05-02
Refactor: GCS Importer

rinickolous:
This PR refactors the GCS Importer Schema functionality for easier support of nested schemas and recursive models (i.e. all item arrays, where items can include children of the same type).


mjeffw (approved):
Fix the merge conflicts and this is good to go.

PR #2696, opened 2026-05-02
Bugfix: modifierTags handling

rinickolous:
`modifierTags` is sometimes `undefined` or a type other than `string`, `Array` or `Set` and is therefore not handled correctly when processing an OTF. This PR properly accounts for the various possible types `modifierTags` could belong to.


mjeffw (approved):
Fix the merge conflicts.

PR #2704, opened 2026-05-03
Fix: Issues with setting posture

### LINKED ISSUE #2699: V1.0.0 GurpsActorV2.toggleStatusEffect fails if changing posture and previous posture is not standing
If you change the posture of an actor (via GurpsActorV2.replacePosture) and its previous posture is not "standing" (so a posture status effect is present)  then it fails in GurpsActorV2toggleStatusEffect when trying to remove the old posture.

Error: 
foundry.mjs:6137 Uncaught (in promise) Error: undefined id [YDtBNkqewEwi7Fy9] does not exist in the EmbeddedCollection collection.
    at EmbeddedCollection.get (foundry.mjs:6137:36)
    at #preDeleteDocumentArray (foundry.mjs:78962:30)
    at async ClientDatabaseBackend._deleteDocuments (foundry.mjs:78932:5)

The reason seems to be that it tries to remove an ActiveEffect, but when a posture is set, on only status is toggled, no Active Effect create.

rinickolous:
This PR slightly refactors the process of changing posture via `GurpsActorV2#toggleStatusEffect`.

A breakdown of changes:
- Moved functions for checking whether a status is a posture status out of `GurpsActor`.
- Cleaned up the embedded document updates and reduced render calls, so the sheet is only re-rendered once when a new posture is selected.
- Made the posture conditions into an enum rather than using string literals
- Moved the posture and maneuver setting events out of the render method and into the (more correct as far as Foundry's API goes) `_onChangeForm` method
- Prevented the ActiveEffect flag update which fires when an ActiveEffect is created from causing associated applications to re-render.


mjeffw (approved):
Fix the merge conflicts.

PR #2721, opened 2026-05-06
fix: ensure damage application works for linked tokens and handle nul…

### LINKED ISSUE #2720: Dropping damage on linked token fails to apply damage

mjeffw:
…l token cases


mjeffw:
@rinickolous By the way, and this is nothing more than a gentle reminder, your earlier change the the DamageCalculator to try to handle both tokens and non-tokens (Actors) in the Apply Damage Dialog broke this functionality for linked tokens while not also making it work for Actors. You may need to check your testing discipline.

Issue #2761, opened 2026-06-10
V1.0.0 Refresh of the Effects Modifier window fails


StefanLeng:
Hitting the Refresh button in the Effect modifier widow results in the following error in the console:

effect-modifier-popout.js:324 Uncaught (in promise) TypeError: actor.applyItemModEffects is not a function
    at EffectModifierPopout.refreshUserMods (effect-modifier-popout.js:324:30)
    at Object.onclick (effect-modifier-popout.js:312:33)
    at HTMLAnchorElement.<anonymous> (foundry.mjs:42338:16)
    at HTMLAnchorElement.dispatch (jquery.min.js:2:40035)
    at v.handle (jquery.min.js:2:38006)

Possibly related to #2710

StefanLeng:
This is fixed by https://github.com/crnormand/gurps/pull/2771

Issue #2770, opened 2026-06-21
V1.0.0 Modifiers from maneuver are not added to User Modifiers


StefanLeng:
In 0.18.x appropriate modifiers are added automaticaly to the actors userModifiers when a maneuver like All Out Attack is selected.
In V1.0.0 this don't happens. May be related to Status: Open.
#2710

StefanLeng:
This is fixed by #2771

PR #2771, opened 2026-06-23
refactor addTaggedRollModifiers

### LINKED ISSUE #2710: V1.0.0 Can not add User Modifiers
If I try to add a user modifier via the effect modifier popup, i get the add dialog, but not added to the effect modifier window when I save it.

Also, on the actor system.conditions.usermods now seems to be derived data and contains a set with the IDs of all Items of the actor, In previous versions this was a persistent array with the modifiers the user entered via the effect modifier popup, not with modifier derived from items.

StefanLeng:
Refactors addTaggedRollModifiers to be more robust an readable,
Can't be fully tested yet, because for much of the functionality to work, I have first to fix #2710 (User Modifiers).

This is a first step. It is a bit absurd, that we basically have all necessary information about the roll in GURPS.performAction and then encode it in some strings and have to decode it here AND look at some other optional data .
If we could make sure that doRoll is for all rolls only called from performAction, it should be possible to work with just the Action Object for all this.
We could also consider storing the tagged modifiers as a structured object instead of encoding tags and source in the string,

I am not sure  how big the scope of this PR should be. I think I would prefer to limit it to this function and fixing the user modifiers and do more later,

Update: now fixes #2710 and #2761 
I opted to store the user created custom modifiers in the holderItem of the actor. This seem to be a appropriate place to me and integrates nicely with the modifiers from regular items.

### DIFF
--- a/src/module/tagged-modifiers/tagged-modifiers.ts
+++ b/src/module/tagged-modifiers/tagged-modifiers.ts
@@ -1,0 +1,220 @@
+import { MeleeAttackModel } from "@module/action/melee-attack.js"
+import { RangedAttackModel } from "@module/action/ranged-attack.js"
+import { ActionType } from "@module/action/types.js"
+import { cleanTags } from "@module/actor/effect-modifier-popout.js"
+import { AnyObject } from "fvtt-types/utils"
+
+import { TaggedModifiersSettings } from "./index.js"
+
+
+enum ROLL_TYPE {
+  MELEE = 'm',
+  RANGED = 'r',
+  PARRY = 'p',
+  BLOCK = 'b',
+  DAMAGE ='d',
+  SKILL = 'sk',
+  SPELL = 'sp',
+  DODGE = 'dodge',
+  ST = 'st',
+  DX = 'dx',
+  IQ = 'iq',
+  HT = 'ht',
+  WILL = 'will',
+  PER = 'per',
+  FRIGHT_CHECK = 'frightcheck',
+  VISION = 'vision',
+  HEARING = 'hearing',
+  TASTE_SMELL = 'tastesmell',
+  TOUCH = 'touch',
+  CR = 'cr',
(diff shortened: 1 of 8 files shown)


src/module/tagged-modifiers/tagged-modifiers.ts
```diff
@@ -0,0 +1,220 @@
...
+    // ToDo: In V1.0.0 the item references from the roll don't match, because the rolls uses the names, and TockenActions use keys.
+    // To fix this I have to fix the UserModifiers first.
+    // This whole logic here is quite fragile, the item matching should use item IDs.  
+    const tokenActionModsFits = !mod.includes('#maneuver') || mod.includes('@man:') || mod.includes(itemRef)
+
+    //check for combat/noncombat
+    const nonCombatModFits =  !actorInCombat || !taggedSettings.nonCombatOnlyTag || !allTags.has(taggedSettings.nonCombatOnlyTag)
+    const combatModFits =  actorInCombat || !taggedSettings.combatOnlyTag || !allTags.has(taggedSettings.combatOnlyTag)
+    
+    //ToDo: the original code has the following code. I have no clear idea how this is supposed to work and I can find on documentation of such a feature
```
StefanLeng, line 179: @mjeffw : Do you have any idea what was the original idea here?

StefanLeng:
As far a I can see, the tagged modifier functionality now fully works as expected. 
This fixes #2660, #2710, #2761, #2770 as well as some issues with status effects I encountered during testing.
 
I still want to add some test for tagged-modifiers.ts, but I have trouble to get vitest to run. I will investigate, but run of time currently.

I will be traveling for the next two weeks without assess to Foundry and can not do any development during that time. I will be back for any fixes or suggestions after that.

token-action.js could benefit from some refactoring (I think this functionality should be on the GurpsCombatant class), but that is out of scope for this PR.

mjeffw (approved):

PR #2787, opened 2026-07-15
fix: misc /li command issues

### LINKED ISSUE #2134: 17.17 The /light command has multiple bugs
1. Running the /light chat command with no arguments produces an error instead of the expected list of animation types.

`Error: /light is not a valid chat message command.`

Image

2. The LightIntensity parameter appears to have been replaced with color intensity, which is not the same thing as light intensity was in earlier versions of foundry. 

3. There is no place to specify light.luminosity, which is a much more important setting in V10+ than color intensity.

rinickolous:
This should address all issues in #2134 

Fixes #2134


rinickolous:
I think we'll probably end up re-doing this command later because quite frankly it's gross right now

mjeffw (approved):

Issue #2794, opened 2026-07-21
V1.0.0 HP and FP can not be updated on the modern sheet


StefanLeng:
On the modern sheet, the value of HP or FP can not be updated.
The modern sheet has just an input with name="system.HP.value". 
system.HP.value is not an updateable value in the V.1.0.0 data model.

PR #2795, opened 2026-07-22
Allow HP/FP above base value

### LINKED ISSUE #2757: V1.0.0 Can not increment HT or FP above the base  Value.
To be able to increment this values above base value is useful for temporary Hit Points or similar effects. It is possible to do so in 0.18 .

StefanLeng:
Modified the data model and the GCS Sheet to allow  for attribute pool values above base value for things like temporary HP.
Fixes #2757 
Should also work with the modern sheet, but that is blocked by #2794

### DIFF
--- a/src/module/actor/data/character-components.ts
+++ b/src/module/actor/data/character-components.ts
@@ -26,7 +26,7 @@ class CharacterPool extends DataModel<PoolSchema> {
 
 const poolSchema = () => {
   return {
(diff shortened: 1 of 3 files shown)


mjeffw:
I've been busy getting ready for Gen Con. I'll address these PRs soon.

StefanLeng:
No worries. I do know you are planing big things for Gen Con, so I did expect no activity here before that. 
Have fun!

mjeffw (approved):

PR #2798, opened 2026-08-03
Fix pseudo document updates on synthetic actors

### LINKED ISSUE #2714: V1.0.0 Can not edit custom recurce tracker
On the GCS Sheet, I can add a custom tracker. In Edit mode I can open the edit dialog for the tracker, but non of the changes are stored when I press update. That is also the case if I try to apply a preset.
On the modern sheet, the behavior for existing trackers is the same, but pressing the + button to add a tracker does nothing.

On both sheets, the custom trackers can not be increased or decreased.

StefanLeng:
fixes #2714

Please review critically, I am quite  sure this fixes the issue, but I am not sure that  I am not just fixing the symptom of a deeper problem.

The issue is that the source of the stored document in the collection does not refer to the source of the parent document in the case of synthetic actor ( and only in this case).  So it doesn't receive the changes from a update and the document is not updated in the collection. 

But I could not find what causes this inconsistency.  So I fixed it in initializeDocument. Once fixed, it stays fixed, so the document is not recreate on ever initialization of the actor, but only on the first per session.

### DIFF
--- a/src/module/data/model-collection.ts
+++ b/src/module/data/model-collection.ts
@@ -235,7 +235,10 @@ class ModelCollection<Model extends PseudoDocument.Any = PseudoDocument.Any> ext
   #initializeDocument(data: AnyMutableObject, options: AnyMutableObject): Model | null {
     let doc = this.get(data._id as string)
 
-    if (doc) {
+    // if the parent is the synthetic actor of an unlinked token, the source of the stored document can be out of sync with the source of the synthetic actor.
+    // I am unclear why this happens.
+    // but in this case we cannot just reinitialize the document, we need to create a new one.
+    if (doc && doc._source === data) {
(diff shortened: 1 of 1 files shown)


mjeffw:
@rinickolous Can you look at this one? I'm not entirely confident with the pseudo document class.

rinickolous:
This fix is correct, it now re-links the `_source` for the `PseudoDocument` where that `_source` belongs properly to the ActorDelta (I think?) rather than the Actor.

However, it exposes another edge case where an unlinked Token's PseudoDocument inherited from the base Actor gets into a bad state if:
1. The unlinked Token's copy has been modified (i.e. there is an ActorDelta entry for it)
2. The PseudoDocument is deleted from the base Actor.

I'm just trying to address that now.

PR #2804, opened 2026-08-06
Fix pseudo document updates on synthetic actors

rinickolous:
This builds off of #2798 by @StefanLeng 

In addition to the fix in that PR, this PR forces PseudoDocuments modified on an unlinked Token do effectively detach from ownership by the base Actor to ownership by the unlinked Token itself. This is accomplished through hydrating the PseudoDocument source object when any field of it is updated AND the Actor is an unlinked Token.

This PR also includes a general type-fix and dependency update pass, though only the following files contain differences relevant to #2714:
- `gurps-actor.ts`
- `collection-field.ts`
- `gurps-actor.test.ts`

I can split the other changes into a separate PR if really necessary, though that would need to be merged before this for this to pass CI.


src/module/actor/gurps-actor.ts
```diff
@@ -80,13 +80,81 @@ class GurpsActorV2<SubType extends Actor.SubType> extends Actor<SubType> {
...
+   * Ensure an unlinked Token's ActorDelta stores complete pseudo-document
+   * sources rather than patches which depend on the base Actor's source.
+   */
+  override async update(
+    data: Actor.UpdateInput,
+    operation: Actor.Database.UpdateOneDocumentOperation = {}
+  ): Promise<this | undefined> {
+    const touchedPseudoDocuments = new Map<string, Set<string>>()
+
+    if (this.isToken) {
```
mjeffw: Ideally, you'd break this function up into separate called functions:
1. Get the list of touched PseudoDocuments
2. Get the list of delta updates
3. Update the deltas

mjeffw:
> This builds off of #2798 by @StefanLeng
> 
> In addition to the fix in that PR, this PR forces PseudoDocuments modified on an unlinked Token do effectively detach from ownership by the base Actor to ownership by the unlinked Token itself. This is accomplished through hydrating the PseudoDocument source object when any field of it is updated AND the Actor is an unlinked Token.
> 
> This PR also includes a general type-fix and dependency update pass, though only the following files contain differences relevant to #2714:
> 
> * `gurps-actor.ts`
> * `collection-field.ts`
> * `gurps-actor.test.ts`
> 
> I can split the other changes into a separate PR if really necessary, though that would need to be merged before this for this to pass CI.

It's *ALWAYS* necessary to split each type of change into its own PR, but I'm not going to enforce it this time. If there was a reason to reject or ask for changes to the pseudodocument changes, we'd also have to reject the changes due to updating the fvtt-types dependencies as well. Tying them together is unnecessary. If you can't make the pseudo document changes because when you started the fix, you updated the dependency, then just submit one PR to update the dependent code, and another with the pseudodocument fixes that depends on the first PR.

mjeffw (changes requested):
Address my one code comment and resolve the merge conflicts and this is good to merge.

src/module/actor/gurps-actor.ts
rinickolous: Donezo

mjeffw (approved):

PR #2843, opened 2026-08-16
Fix technique name handling to avoid "[object Object]" string literal…

### LINKED ISSUE #2781: Techniques include "[object Object]" string literal since recent GCS update
Transferred from: https://github.com/richardwilkes/gcs/issues/1071, originally submitted by @eloisepasteur
Thread transcript follows:

Image
See image. This behaviour is new, since one of the recent GCS updates. It should say something like Feint (weapon) or I think the wrestling targeted attack is complete, but adds an extra (object Object) anyway.

richardwilkes responds:
OK, after looking at this in GCS, I can't replicate anything that would produce such output in the saved file. Please zip up the .gcs file and attach it here so I can take a look (if you don't zip it, GitHub will refuse to let you attach it). I suspect I won't find anything in that file, either, but it will let me rule it out. Assuming that's the case, you'll need to talk to the GGA developers about the issue instead.

eloisepasteur responds:

[Harriet.Atkinson.gcs.zip](https://github.com/user-attachments/files/29978002/Harriet.Atkinson.gcs.zip)
This isn't the same character sheet as the screen grab is from (that was someone else in the same game and I used their screen grab, not mine).

If you're going to search for things, Cmd+F style, this is the top of my character sheet and shows a few examples for you.

Image

richardwilkes responds:
Thanks. The data file shows that GCS isn't accidentally emitting those "[object Object]" bits in its calc sections that it emits for other programs to use. This means that the problem is somewhere in the GGA code, not in GCS. @rinickolous if you'd like me to transfer it to some other repo, let me know the destination and I'll transfer it -- assuming GitHub will let me do so. Otherwise, I'll close the issue.

eloisepasteur responds:
Thanks for checking. Sorry for bothering you. After the recent updates to GCS, it seemed more likely the problem was with that output.

mjeffw:
… in GCS import

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -19,6 +19,8 @@
   - After installing the update, go to the settings for `GURPS 4e Game Aid > PDF Settings`, and change `Basic Set PDFs` to "Basic Revised ('B')".
 - Detached combat tracker window/popout! window not allowing updates to maneuvers. \#2796
 - GCS v5 import: spell points inside containers are not summed (+ minor: importSp never sets spl.level) \#2772
+- Techniques include "[object Object]" string literal since recent GCS update \#2781
+  - You may need to delete the techniques and reimport to apply the fix.
 
 ## Release 0.18.20 06/18/2026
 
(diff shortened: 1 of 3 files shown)


eloisepasteur:
Not fixed for me. I have just tried importing a character and I get this:
Screenshot 2026-08-16 at 20 37 24

mjeffw:
@eloisepasteur You have to wait until the next release.

eloisepasteur:
Ah, sorry. I was very keen to see it

mjeffw:
@eloisepasteur Just released today!

PR #2876, opened 2026-08-23
Fix quick roll menu vor 1.0.0

### LINKED ISSUE #2868: V1.0.0 Combat tracker Quick roll menue broken
No checks beside the attribute checks are rendered, because the struktur of the data returned from actor.getChecks is mismatched with the handlebar template.
The CSS is broken, making the rendered text illegible.

StefanLeng:
Fixes #2868 
There was a lot of stuff broken. I did the following fixes:

- character.getChecks returned a different structure than in V0.18. I mostly keep the V1.0.0 structure and adapted the HBS, as the v0.18 structure was needlessly complicated.
- many of the OTFs returned by character.getChecks where wrong and some other data was missing or wrong. I tried to use values from the display items where possible, to avoid duplication of logic.
- character.getChecks did not honor the addToQuickRoll Flag of attacks.
- The addToQuickRoll flag of attacks was missing in the action edit sheet.
- character.getChecks did not return entries for OTFs of selected items. 
- The menu did not honor the UI Theme and was partially unreadable in light mode. Also fixed this for the maneuver dropdown.
- The switch from attack to damage rolls on CRLT did not work.
- I tried  for a slightly more compact and consistent layout.
-  I converted the file to TS and fixed  the Type errors.
- I removed JQuery in favor of plain JS

There is a non functional, invisible Blind Roll switch in the HTML and some of the code. I like to remove it, as it doesn't make sense to have a blind roll switch as a global switch exists. This ties this code to the TokenActions, where the state of the switch is stored. Removing it would reduce code complexity and dependencies. What do you think?

### DIFF
--- a/src/module/actor/data/character.ts
+++ b/src/module/actor/data/character.ts
@@ -1498,6 +1498,7 @@ class CharacterModel extends BaseActorModel<CharacterSchema> {
                 value: displayItem.level,
                 mode: displayItem.usage,
                 otf: displayItem.otf.level,
+                damage: displayItem.damage,
                 isOTF: true,
                 otfDamage: displayItem.otf.damage,
                 type: attack.type === 'meleeAttack' ? 'melee' : 'ranged',
--- a/src/module/actor/data/character.ts
+++ b/src/module/actor/data/character.ts
@@ -11,12 +11,13 @@ import { ModelCollection } from '@module/data/model-collection.js'
 import * as HitLocations from '@module/hitlocation/hitlocation.js'
 import { BaseItemModel } from '@module/item/data/base.js'
 import { ConditionalModifier, ReactionModifier } from '@module/item/data/conditional-modifier.js'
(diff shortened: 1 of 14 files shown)


mjeffw:
> There is a non functional, invisible Blind Roll switch in the HTML and some of the code. I like to remove it, as it doesn't make sense to have a blind roll switch as a global switch exists. This ties this code to the TokenActions, where the state of the switch is stored. Removing it would reduce code complexity and dependencies. What do you think?

Yes, go ahead and remove the nonfunctional switch. 

I'm not sure that a global switch is the right solution but that's another issue. I generally disfavor global variables.

StefanLeng:
Oh, the switch is just the control foundry provides near the chat window. The user can set the message mode there. So we don't need another control to set the message mode in this window.
global was just a bad wording in my message.

StefanLeng:
non functional switch removed

mjeffw (approved):

PR #2893, opened 2026-08-31
Add GM setting to toggle maneuver modifiers in the Modifier Bucket

ncphillips:
## Why

GURPS ships far more maneuvers and attack/defense options than most campaigns run with. On Target adds more still, and Martial Arts options like Telegraphic Attack or Deceptive Defense are optional rules a table may never have adopted. Until now every one of them was offered to every player in every world: the maneuver list was fixed, and the Modifier Bucket showed all 32 options whether or not the group used them.

That is a lot for a GM to talk their table past, and it does not go away — the options a campaign has never adopted sit in front of the players every time they attack or defend.

## What

A **Combat Options** dialog where a GM says what the campaign is actually using: the On Target source toggle, the maneuvers in play, and the attack/defense options offered in the Modifier Bucket.

A maneuver that is turned off disappears from everywhere a maneuver can be picked — the sheet dropdown, the token HUD, the combat tracker menu and `/man` — while still resolving correctly on any token already using it. An attack or defense option that is turned off leaves the bucket, which now also drops a section heading rather than leaving it stranded over nothing once every option beneath it is gone.

### Maneuvers and modifiers are coupled, not independent

A modifier that describes a maneuver is meaningless once that maneuver is gone: turning off All-Out Defense (Dodge/Parry/Block) takes the "+2 to defense (AoD: Increased)" line with it. The dialog shows that cascade instead of applying it invisibly — a row that cannot reach the bucket is disabled and says which maneuver, or which source, it is waiting on.

Checkbox state is read from the DOM rather than from the form data on submit, so a disabled row still saves and a GM's choices survive turning a maneuver off and back on.

Do Nothing and Move are not offered at all. Do Nothing is what the system falls back to when an actor has no maneuver, hardcoded as such in several places, so a world without it in play would leave a player no way to pick it back.

### Existing worlds

Absent keys count as enabled, so existing worlds start with everything in play and options added in later releases default to visible. The four maneuver settings that moved out of `lib/miscellaneous-settings.js` into the combat module keep their existing keys, so there is nothing to migrate.

## Groundwork

Three changes that are not user-visible but are what the feature rests on:

- The 32 attack/defense options were inline in the Modifier Bucket's rendering code. They are now a registry the bucket reads, which is what makes them filterable at all.
- The maneuver registry and its settings moved into `module/combat/`, next to the code that reads them.
- `getAll()` returned a source-filtered view of the registry rather than the registry. Resolving a maneuver already applied to a token and listing the maneuvers a player may pick are now separate calls, because they need different answers: a maneuver turned off underneath an actor still has to render its icon, label and move.

## Bugs fixed along the way

Two of these are defects in this PR's own new code, found while reviewing it. The third was already on `main`.

**Introduced here, fixed here.** Both live in `refreshCombatOptionUI`, which is new in `b54ed1ce`:

- *Refreshing the UI opened every actor's sheet.* Finding the open sheets by asking each actor whether `actor.sheet.rendered` is what opened them — `Document#sheet` is a lazy getter that constructs and caches an Application on demand. Every combat option change built a sheet object per actor in the world, on every connected client.
- *One Save rebuilt the UI twice.* Saving writes two settings, and the first rebuild showed a state nobody asked for — On Target off, its maneuvers still listed — before the second corrected it. The refresh is now debounced.

**Already on `main`, fixed here.** `getManeuver()` looked a stored maneuver id up only among the source books in use and dereferenced without a guard, so switching On Target off under a token mid Committed Aim made that actor's sheet fail to render. `getAll()` was itself the filtered set before this PR renamed it, and On Target has always been togglable, so the crash predates the branch — but the dialog made it easy to reach, since changing an option now re-renders every sheet immediately.

## Reviewing this

**Commit by commit is much easier than the squashed diff.** The refactors are separated from the behavior changes, and every one of the twelve commits builds and passes the suite on its own — `npm test` and `npx tsc --noEmit` are clean at each.

**Happy to split this into a series of PRs if the core team would rather.** The natural seams are the three groundwork commits (registry extraction, the module move, the `getAll()` split), then the dialog itself, then the fixes above. Say the word and I'll break it up.

## Screenshots
image

image

image

image

image

#### Modifier Popup
image
#### Token Maneuver Selection
image


lang/en.json
```diff
@@ -1356,6 +1358,9 @@
     "settingManeuverGMOnly": "GM and Owner Only",
     "settingManeuverMove": "Maneuver Updates Move",
     "settingManeuverNoOne": "No One",
+    "settingManeuverOptions": "Bucket: Maneuvers",
```
ncphillips: Maybe something more like

- "Maneuvers In-Play"
- "Available Maneuvers"
- "Active Maneuvers"

Quite open to suggestions

lib/miscellaneous-settings.js
```diff
@@ -455,6 +458,30 @@ export function initializeSettings() {
...
 
+    game.settings.register(SYSTEM_NAME, SETTING_MANEUVER_OPTIONS, {
+      name: game.i18n.localize('GURPS.settingManeuverOptions'),
+      hint: game.i18n.localize('GURPS.settingHintManeuverOptions'),
+      scope: 'world',
+      config: false,
+      type: Object,
+      default: defaultManeuverOptionSettings(),
+      onChange: value => {
+        console.log(`Maneuver options: ${JSON.stringify(value)}`)
```
ncphillips:
```suggestion
```

mjeffw:
At first glance, a well-constructed PR. Especially love the unit tests, thanks!

I think that instead of living in `module/modifier-bucket`, that code could be moved to `module/combat` and function(s) of the module be exposed for the modifier-bucket to read.

Also, and this is optional -- moving the settings from `lib/miscellaneous-settings` into the `combat` module would be the way to go for our future-looking architecture. That would mean getting all the other maneuver settings and moving them there as well. Look at `module/pdf/settings.ts` as an example. NOTE: This is purely optional because it increases the scope of this PR, so only do it if you're really motivated to do so. LOL

ncphillips:
Thanks! I'm doing some heavy adjustments actually. There's some things that I'm not loving. I'm down with increasing the scope a bit.

StefanLeng:
Should this not also modify the maneuver selection on the character sheets?

lang/en.json
mjeffw: "Available Maneuvers" is the best wording IMO.

mjeffw:
> Should this not also modify the maneuver selection on the character sheets?

Yes, it should. One possibility is to filter the maneuvers directly in `module/actor/maneuver.js:Maneuvers.getAll()` and ensure all code reads the maneuvers from this function.

And this file `maneuver.js` should also be moved to the `combat` module.

ncphillips:
Greatly appreciate your notes 👍 FYI I'm force pushing as I work on this. I'll mold the PR a bit to have some clear commits with detailed messages.

ncphillips:
Opening up now. I would recommend review commit-by-commit. I arranged things so each commit builds and has fairly detailed messages. A bug fix and some refactorings are at the start, the feature itself is at the end. 

It's a bit much, so if yahs would prefer I'd be happy to split it into multiple PRs.

**Update:** Actually, this is a bit much. I'm going to split things out a bit.

mjeffw:
> **Happy to split this into a series of PRs if the core team would rather.** The natural seams are the three groundwork commits (registry extraction, the module move, the `getAll()` split), then the dialog itself, then the fixes above. Say the word and I'll break it up.

Nice, I much prefer multiple PRs to one big one. Splitting it up would be great. Thanks.

Otherwise, is this ready for review?

ncphillips:
I believe so! You can pull this if you want to get a feel for the whole thing. I've closed this PR entirely though. 

The first two PRs are open now #2897 and #2898 

I'll have one more refactoring after them and then the feature itself.

PR #2933, opened 2026-09-14
Extract confirm roll dialog to own appliaction V2

StefanLeng:
This extract the Roll Confirmation Dialog into its own AppV2.

It this process, I converted dieRoll.js to typeSkrip. This is just the first step, I will refactor this code in a further PR.

I removed GURPS.handleRoll because I always expect an action object in doRoll. It is not called in the code base anymore. If you think its important for third party code to retain this function, it would have to be rewritten to translate all calls to performAction or executeOTF. That can be done, but honestly i an not sure i understand all code paths of handleRoll correctly.

I also changed all calls of GURPS.performAction, GURPS.executeOTF and GURPS.parslink to the GURPS.modules.OTF interface.


rinickolous:
man, really saving me a lot of work I was dreading doing.

src/module/util/allow-otf-exec.js
mjeffw, line 1: This file should probably be moved to the OTF module and converted to TS.

src/module/gurps-wiring.js
mjeffw, line 1: Do you think this should be moved to the OTF module?

mjeffw (changes requested):
Just two minor comments.

src/module/gurps-wiring.js
StefanLeng: Yes, I think so. But I like to handle some other stuff first, manly trimming down and standardizing the interface between the actionFuncs and doRoll and the tagged modifiers. But can tackle this file later.

src/module/util/allow-otf-exec.js
StefanLeng: I agree. I will move and convert it as part of my next PR. I think this one is big enough as it is, so I like to handle it in the next.

PR #2934, opened 2026-09-14
Bugfix/Feature: #2774 Edit Action for Active Effects (v1.0.0)

### LINKED ISSUE #2774: V1.0.0 No way to edit active effects
None of the V1.0.0 Character sheets offer a way to edit the active effects of a character,

rinickolous:
This PR adds edit functionality to all 3 1.0.0 actor sheets.

## Summary of Changes

- Added "Add Effect" button to GCS Sheet, mimics Modern Sheet functionality
- Moved "Active Effects" from the Maneuver / Status to its own section due to limited space in the former (GCS Sheet)
- Made the Active Effects list in the GCS sheet actually display the active effects list, similarly to the Modern Sheet.
- Added an "Edit This Effect" button for active effect entries on all 3 sheets.
- Simplified the active effect display markup on the Modern Sheet.
- Removed the `deleteEffect` action from Modern Sheet (superseded by universal `deleteEmbedded`)


mjeffw:
Something wrong with this PR -- in the original PR when I hover the mouse over the AE, the Edit control appears. It no longer does so on either Modern or GCS sheet.

Also, the way that was implemented caused the AE control to change size horizontally. I don't like that behavior, because when you have a wrapped list of AEs, the change in size may change where the list if wrapped. Imagine if you have two lines of AEs, and you want to edit the last one of the first line. Hovering over it and increasing the size may make the AE wrap to the next line. It may result in never being able to edit that effect.

I see two possible solutions -- one is the reserve space for the edit control (in which case you might just show it without hovering) unless you have the edit button appear over the last bit of text on the AE control. (Does that make sense?)

The other solution may be to make the text of the AE widget trigger editing (i.e., you click on the text and it opens the editor). In that case I think we'd need some visual cue that this is the behavior... (highlighting on hover?).

Finally, we need a way to create custom AEs. I think you could put a button in the Effect Picker to the right of the search bar:

image
 
This should create an AE with generic data and open the editor on it.

mjeffw (changes requested):
See my comments in the Conversation section.

rinickolous:
> I see two possible solutions -- one is the reserve space for the edit control (in which case you might just show it without hovering) unless you have the edit button appear over the last bit of text on the AE control. (Does that make sense?)
> 
> The other solution may be to make the text of the AE widget trigger editing (i.e., you click on the text and it opens the editor). In that case I think we'd need some visual cue that this is the behavior... (highlighting on hover?).

Here's a proposal: Move both Edit and Delete functions to a context menu which can be activated by either right-clicking on any part of the AE widget, or by clicking on a vertical ellipsis / kebab button (which replaced the AE's image on hover). I would make sure the image replacement does not impact the widget's size, of course. This means less negative space which would otherwise be used as space reserved for the Edit and Delete icons. Let me know what you think @mjeffw .

mjeffw:
I want to maintain "discoverability" -- i.e., the user can see that they can interact with this widget without randomly (right-)clicking on stuff. The context menu attached to an ellipsis or kebab button (LOL, I had to look that up) that appears where the AE image is on hover makes the most sense to me.

Also add an enable/disable option to that menu. A disabled AE should be displayed differently, for example in a strikethrough text, or greyed out, or both.

PR #2939, opened 2026-09-14
Bugfix: #2895 GCS Sheet Displays Unadjusted Attribute Values

### LINKED ISSUE #2895: V1.0.0 GCS Actor sheet doesn't show Attribute Values adjusted by Active Effects
If you have a Active Effect that e.g. adjusts system.attributes.ST.value +5, this change shows up in the Modern Sheet and also when you do a ST Roll.
But on the GCS Sheet the unadjusted value is shown (actually the display here is importedValue instead of Value).

I think in Play mode, the adjusted value should be shown. In Edit Mode the unadjusted value should be shown, to avoid accidentally saving the adjusted value. Usually, that should be the corresponding value in system._source but in this case it seems to be system._source.attributes.ST.importedValue.

rinickolous:
This PR fixes the GCS sheet's attribute display so that it displays the post-bonus value in play mode.


src/module/actor/sheets/gcs-actor-sheet.ts
```diff
@@ -286,82 +287,98 @@ class GurpsActorGcsSheet extends GurpsBaseActorSheet<
     const systemSource = this.actor.system._source
     const attributeFields = this.actor.system.schema.fields.attributes.fields
     const attributeSource = this.actor.system._source.attributes
+    const attributeInst = this.actor.system.attributes
```
mjeffw, line 290: What does "attributeInst" mean? I suggest spelling out the variable name instead of abbreviating it.

mjeffw (changes requested):

mjeffw:
I still want `attributeInst` disambiguated.

PR #2946, opened 2026-09-16
Combat Settings: Maneuvers + Modifiers

ncphillips:
Raised on Discord: https://discord.com/channels/228546929494065152/1324122623403102324/1543987584731058226

## The problem

GURPS ships far more maneuvers and attack/defense modifiers than any one campaign runs with. On Target adds more still, and Martial Arts rules like Telegraphic Attack are optional. Until now every one of them was offered to every player in every world: the maneuver list was fixed, and the Modifier Bucket showed all 32 modifiers whether or not the group had ever heard of them.

## The shape of the solution

Combat Settings gains two lists — the maneuvers this campaign uses, and the stand-alone Modifier Bucket options it offers.

A maneuver that is turned off disappears from everywhere one can be picked: the sheet dropdown, the token HUD, the combat tracker menu and `/man`. A modifier that is turned off leaves the Modifier Bucket, which also drops a section heading once nothing is left under it.

Underneath, the enabling change is a seam that didn't exist: `Maneuvers.getAll()` was serving two unrelated questions. *What may a user choose?* (the pickers) and *what is this maneuver?* (resolving an id already stored on an actor into its label, icon and move). Narrowing the choosable set would silently have narrowed resolution too, which is the wrong answer — a token mid Committed Aim keeps its maneuver even after the GM turns it off. So `getAllInPlay()` is now the picker list and `getAll()` the resolution list.

### Maneuvers
image

### Modifiers
image

## Detail

**The two lists are not independent.** A modifier that *defines* a maneuver (+4 to hit **is** All-Out Attack (Determined)) has no switch of its own: it follows the maneuver, and isn't listed in the modifier section at all. Only the stand-alone options — Telegraphic, Deceptive, Rapid Strike, Acrobatic Dodge — get a checkbox. `COMBAT_OPTIONS` records which is which via a new `maneuvers` field, and a stored value for a defining modifier is ignored rather than obeyed.

**Do Nothing and Move are always in play.** Do Nothing is what the system falls back to when an actor has no maneuver at all; Move is the baseline the others are described against. Turning either off left a player with no way back and no way to tell why, so they aren't offered, and a settings object that turns them off anyway is overruled.

**Absent keys count as enabled.** Existing worlds start with everything in play, and anything added in a later release defaults to visible rather than silently off. No migration.

**Changing the setting re-renders what's on screen.** The Modifier Bucket reads lazily but may already be open; the tracker menu, HUD palette and sheet dropdowns are each built once per render. `refreshCombatOptionUI()` re-renders all four. Turning On Target on or off now goes through the same path, since it also adds and removes maneuvers.

**Three bugs fixed along the way,** all of which predate this branch:

- `getManeuver()` did `fromSourcesInUse()[id].data` with no guard, against the set of maneuvers from source books *currently in use*. Being applied to a token and being offered by a source book in use are independent — a GM can switch On Target off while a token is mid Committed Aim. When that happens the lookup misses, `.data` throws, and since the `getManeuver` Handlebars helper runs during sheet render, **the actor becomes unopenable** until the setting is put back. It goes unnoticed because the throw arrives late, the next time somebody happens to open that particular actor. Resolution now falls back past the in-use set to the full registry, and past that to Do Nothing with a warning.
- The token HUD and the sheet read `maneuver.icon`, which has never been a property — it's `img`. Both were showing `man-nothing.png` for every maneuver. (`518c5806`, `904f49cc`)
- `GurpsSettingsApplication` renders anything that isn't a Boolean, Number or DataField as a text input and saves it back as a string. It now skips Object and Array settings, which have dialogs of their own — otherwise the new setting would have been saved as `"[object Object]"` by any other module's settings dialog.

The six commits are meant to be read in order; each one is self-contained and the messages carry most of the reasoning above.

## Areas that might be non-obvious

- **The dialog reads its checkboxes back out of the DOM on submit, not out of the form data.** The maneuver and option checkboxes are deliberately unnamed. A disabled input submits nothing, and the On Target rows are disabled while On Target is off — reading the DOM is what lets a GM turn On Target off and back on and find their choices intact.
- **The On Target rows are always listed, disabled and struck through with a tooltip when that source is off,** rather than hidden. Ticking Use On Target in the same dialog lights them up without saving and reopening. I went back and forth on this; hiding them is also defensible.
- **`renderedActorSheets()` reads `actor._sheet`, not `actor.sheet`.** `Document#sheet` is a lazy getter that *constructs and caches* an Application for any actor that hasn't got one, so asking every actor in the world whether its sheet is open would be what opened them. It also walks canvas tokens, because `game.actors` misses the synthetic actors behind unlinked tokens — the usual case for mooks.
- **`getAllInPlayData(keep)`** keeps an actor's current maneuver in the sheet dropdown even after it's turned off, because a `<select>` whose value matches no `<option>` displays the *first* one — "None" — while the active effect, token icon and defense penalties are all still in force.
- **Aim appears exactly once in `getAll()`, with its Basic Set data.** There's no single set holding both variants; callers resolving a maneuver already on a token go through the private helper, which picks the On Target one when that book is in use (it moves half instead of step).
- **The dialog class lives in its own file rather than in `settings.ts`.** The menu registration needs the dialog, the dialog needs the maneuver registry, and the maneuver registry needs the settings. Keeping that one edge out of `settings.ts` is what leaves the module's imports acyclic.

## Feedback welcome — genuinely

I'm new enough to this codebase that I'd rather be corrected than merged. Please push back on any of it; I won't be disappointed, and I'm happy to rework or drop pieces:

- **Scope.** Is per-modifier configurability wanted at all, or is the maneuver list the part people actually asked for? The modifier list is the easiest half to cut.
- **The maneuver/modifier coupling.** Making defining modifiers follow their maneuver rather than have their own checkbox is a judgement call. If a GM might plausibly want All-Out Attack (Determined) available as a maneuver but not as a bucket entry, that's the wrong call.
- **Where this lives.** It's under Combat Settings with the rest of the combat module's settings. It could just as well be its own menu.
- **UI.** Layout, wording of the hints, whether the maneuver icons earn their place, whether there should be select-all/none buttons for lists this long.
- **`ALWAYS_IN_PLAY`.** Is `['do_nothing', 'move']` the right set? Should Attack be in there too?
- **Naming.** `getAllInPlay` / `getAll` / `getAllInPlayData` is a mouthful. Better names welcome.
- **Test style.** I've written these in the style I'm used to (one assertion, description reads like the assertion); if that clashes with the project's conventions, say so and I'll match.


lang/en.json
```diff
@@ -1271,6 +1271,11 @@
     "settingColorSheetMenuHint": "Color Character Sheet Settings",
     "settingColorSheetMenuTitle": "Color Character Sheet",
     "settingCombatInitiative": "Combat: Turn Sequence Formula",
+    "settingCombatOptions": "Combat Options",
+    "settingCombatManeuversHint": "A maneuver that is turned off disappears everywhere one can be picked -- the combat tracker menu, the token HUD and the character sheet -- and takes the modifiers that define it out of the Modifier Bucket with it.",
+    "settingCombatModifiersHint": "The optional modifiers the Modifier Bucket offers. One that is turned off is no longer listed there; a player can still type it in.",
```
mjeffw: Suggested rewording:

"The optional modifiers presented by the Modifier Bucket. If turned off, it is no longer listed in the Modifier Bucket. The modifier can still be entered manually."

lang/en.json
```diff
@@ -1301,6 +1306,7 @@
     "settingHintCalculateEnc": "If checked, the Encumbrance level will be set automatically based on the carried equipment (and you cannot change the Encumbrance level manually). If this is turned off, you can change the Encumbrance level by clicking on the new level.",
     "settingHintCheckConditionalModifiers": "If checked, the system will look for Conditional Modifiers in the Actor Sheet and add them to Effect Modifiers Popout.",
     "settingHintCombatInitiative": "This is the formula used by Foundry to determine turn order in combat. Warning: Only change this if you know what you are doing! We do not parse this for correctness. Set this field blank to reset the formula back to the GURPS Combat Turn Sequence, B362.",
+    "settingHintCombatOptions": "Choose which combat options this campaign uses. A maneuver that is turned off disappears everywhere a maneuver can be picked -- the combat tracker menu, the token HUD and the character sheet -- and takes its attack and defense modifiers out of the Modifier Bucket with it.",
```
mjeffw: Suggest:
"Choose the combat options used by this world. A maneuver that is turned off is no longer shown anywhere a maneuver can be picked — Combat Tracker, Token HUD, and character sheet — and its modifiers removed from the Modifier Bucket."

module/combat/combat-options.ts
```diff
@@ -60,13 +86,71 @@ export const COMBAT_OPTIONS: readonly CombatOption[] = [
...
+}
+
+/**
+ * The maneuvers with no meaningful "off": Do Nothing is what the system falls back to when an actor
+ * has no maneuver at all, and Move is the baseline every other maneuver is described against. The
+ * Combat Options dialog doesn't offer them, and a settings object that turns them off anyway --
+ * written by an older build, or by hand -- is overruled rather than obeyed, because a world with no
+ * Do Nothing in play gives a player no way to pick it back.
+ */
+export const ALWAYS_IN_PLAY = ['do_nothing', 'move']
```
mjeffw, line 117: I'm okay leaving 'attack' out of this list. I can imagine (strange) campaigns where attack is an invalid maneuver.

module/combat/combat-options.ts
```diff
@@ -60,13 +86,71 @@ export const COMBAT_OPTIONS: readonly CombatOption[] = [
...
+
 export function enabledOptions(
   section: CombatOptionSection,
+  settings: CombatOptionSettings,
   { useOnTarget }: { useOnTarget: boolean }
 ): CombatOption[] {
-  return COMBAT_OPTIONS.filter(option => option.section === section && (useOnTarget || !option.requiresOnTarget))
+  return COMBAT_OPTIONS.filter(
+    option =>
+      option.section === section && (useOnTarget || !option.requiresOnTarget) && isOptionEnabled(option.id, settings)
```
mjeffw, line 154: At some point I'd like "On Target" to be like a Compendium or something that users can add if they want to use them, instead of being hardcoded. Not really a comment for you.

test/combat-option-refresh.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  return actor
+}
+
+const world = ({ actors = [] as any[], tokens = [] as any[] } = {}) => {
+  ;(globalThis as any).game.actors = actors
+  ;(globalThis as any).canvas = { tokens: { placeables: tokens } }
+}
+
+describe('renderedActorSheets', () => {
+  it('returns the sheet of an actor whose sheet is open', () => {
```
mjeffw, line 25: I'm good with this style of test naming. Back when I was a working programmer/technical coach, I was pushing BDD style tests to my teams.

mjeffw (commented):
Honestly, there's not any substantive changes I could identify, I'm just giving you a chance to read my comments. Let me know when you're good with my merging this.

mjeffw:
For each of these PRs, please add a bullet item to `changelog.md` under `Release 0.18.24`. Thank you!

ncphillips:
Thank you @mjeffw! Changes applied and (force) pushed 🎉

mjeffw:
@ncphillips Is this feature complete? Ready for release?

PR #2950, opened 2026-09-17
2948 v100 rewrite if chat processor to use ast parser remove xregexp library

mjeffw:


ncphillips:
@mjeffw this is neat PR. Huge negative diff! 🙌 I'm here with some unsolicited review. I just thought I'd chuck Fable 5.1 at the PR to see what it found:
<details><summary>Fable 5.1 Review Findings</summary>
Code review of PR #2950 (/if chat processor rewrite to AST parser, removes XRegExp) finished. 7 confirmed correctness findings plus one cleanup, ranked by severity:

1. src/module/chat/if.ts:72 – Non-check or unparseable condition returns false after printing the error, so the else branch runs. /if [+1 bonus] [/hp +1d] /else [/hp -1d] prints "must be a check" and then applies /hp -1d. Old code ran nothing. Existing tests only use then-only forms so they pass.
2. src/module/chat/if-block-parser.ts:105 – Curly-brace format never consumes /else between blocks, though the doc comment and help advertise it. /if [DX] {/r 3d6} /else {/say Miss} throws "Unexpected trailing input".
3. src/module/chat/if-block-parser.ts:323 – isCritFormat only matches glued cs:{. A space before the brace (cs: {Critical!}) demotes the whole tail to literal text posted to chat on success. {Hit} cf:{Fumble} throws.
4. src/module/chat/if-block-parser.ts:335 – parseText stops at any embedded /if in a brace body, breaking backslash-separated multi-command blocks with nested /if (advertised in changelog) and prose containing "/if".
5. src/module/chat/if-block-parser.ts:128 – Unbraced nested /if [DX] /if [IQ] [A] [B] fails with trailing-input error, and { /if ...} with leading whitespace fails since parseBlock only detects /if as the first char after {.
6. src/module/chat/if.ts:19 – IfBlockParser.parse is outside process()'s try/catch, so every ParseError escapes to ChatProcessors.handle as a raw error toast with char offsets. Trailing text after [then] (previously tolerated) now hard-fails. Fix: move parse inside the try and report via chatUnrecognizedFormat.
7. src/module/chat/if-block-parser.ts:52 – Nested node's line is built from input.slice(pos) to end of input, so nested checks echo the outer block's trailing text in the private message. Cosmetic.
8. src/module/chat/dsl-parser/if-parser.js:1 – 651-line second parser with zero importers, self-described as unreviewed and unintegrated, uses CommonJS in a type: module package, and gets emitted into dist. Recommend dropping it (and moving if-command.md out of src/) or landing it in a follow-up that wires it in.
</details>

mjeffw:
> @mjeffw this is neat PR. Huge negative diff! 🙌 I'm here with some unsolicited review. I just thought I'd chuck Fable 5.1 at the PR to see what it found:
> 
> Fable 5.1 Review Findings

Thanks, there were a couple of serious bugs caught by that review. Now, of course, I've rewritten the parser again, so it is all new code.

mjeffw:
@rinickolous No comments?

rinickolous:
> @rinickolous No comments?

Will get to this, just started classes

PR #2954, opened 2026-09-20
Fix Move rounding

ncphillips:


ncphillips:
@mjeffw I saw you merge so I opened. TBH haven't had a good chance to really double check things but I think the basic premise is solid. Tried to get as many sourcebook references as possible for behaviour changing claims

This actually came out of me trying to fix the same thin in gurps-hud https://github.com/ncphillips/gurps-hud/pull/28

The original issue was brought up in Discord by elijah https://github.com/ncphillips/gurps-hud/issues/27

module/actor/move.ts
mjeffw: This is great. The long term goal is to take GURPS rules algorithms and save them in a "rules" module. This allows me to grab this file and move it intact.

